PR #3, opened 2020-01-02
Array tuple set syntax

anzboi:
This PR changes the syntax for tuples and sets. These collection types now conforms to the intended spec.

* Sets: `{}`
* Tuples: `()`
* Arrays: `[]`

Use of `()` for tuples now clashes with the classical use of parens to wrap expressions. This PR includes a hack workaround to continue supporting `(<expr>)`.

Includes:
* Syntax parse changes
* `String()` methods for output
* Hopefully does not break `(<expr>)`.
* Updated tests related to the syntax changes and `String()` methods.

No new tests have been added

### DIFF
--- a/syntax/lex.go
+++ b/syntax/lex.go
@@ -187,12 +187,12 @@ func NewLexerWithPrefix(prefix *bytes.Buffer, reader io.Reader) *Lexer {
 
 func (l *Lexer) copy() *Lexer {
 	// Copy most fields
-	new := *l
+	newL := *l
 
 	if l.stack != nil {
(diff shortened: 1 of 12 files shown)


syntax/lex.go
```diff
@@ -184,6 +185,25 @@ func NewLexerWithPrefix(prefix *bytes.Buffer, reader io.Reader) *Lexer {
 	}
 }
 
+func (l *Lexer) copy() *Lexer {
+	// Copy most fields
+	new := *l
```
marcelocantos: nit: `new` is a built-in function. Better to avoid.

PR #4, opened 2020-01-03
Json translator

anzboi:
This PR adds a json translation command to arrai.

Usage:
```bash
# json subcommand reads from stdin
$ echo '<json blob>' | arrai json
```

This functionality is also available in code through the `translate` package

```go
val, err := translate.JSONToArrai(data interface{})
```

Assumes raw json has already been marshalled into an interface. Will error out if `data` is a strongly typed struct (or anything that doesn't come out of `json.Unmarshal()` on a blank interface variable)

### DIFF
--- a/translate/json.go
+++ b/translate/json.go
@@ -1,0 +1,59 @@
+package translate
+
+import "github.com/arr-ai/arrai/rel"
+
+// JSONToArrai translates an object unmarshalled from json into an array value
+//
+// translation follows the rules
+//
+//     object -> {|@,@item|, |key,val|, ...}
+//     array  -> array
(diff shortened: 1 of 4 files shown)


translate/json.go
```diff
@@ -0,0 +1,59 @@
+package translate
+
+import "github.com/arr-ai/arrai/rel"
+
+// JSONToArrai translates an object unmarshalled from json into an array value
+//
+// translation follows the rules
+//
+//     object -> {|@,@item|, |key,val|, ...}
```
marcelocantos: Should be `(key: val, ...)`.

PR #39, opened 2020-02-26
Add AsString() to Set

### LINKED ISSUE #36: Empty string is a *rel.genericSet instead of rel.String
Doing
```shell
$ arrai e '""'
{}
```

Internally an empty `*rel.genericSet` can not be converted to `rel.String`

@marcelocantos suggested `Set.AsString() (string, bool)` method

nofun97:
Add AsString() to Set interface which converts any Set to the internal String struct

Fixes #36

### DIFF
--- a/rel/value_set.go
+++ b/rel/value_set.go
@@ -328,7 +328,7 @@ func (s *genericSet) Any() Value {
 }
 
 func (s *genericSet) AsString() (String, bool) {
-	if s.set.Count() == 0 {
+	if !s.Bool() {
 		return String{}, true
 	}
 
--- a/rel/value_set.go
+++ b/rel/value_set.go
@@ -327,6 +327,55 @@ func (s *genericSet) Any() Value {
 	panic("Any(): empty set")
 }
 
+func (s *genericSet) AsString() (String, bool) {
+	if s.set.Count() == 0 {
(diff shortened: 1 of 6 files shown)


rel/value_set.go
```diff
@@ -327,6 +327,48 @@ func (s *genericSet) Any() Value {
 	panic("Any(): empty set")
 }
 
+func (s *genericSet) AsString() (String, bool) {
+	if s.set.Count() == 0 {
```
marcelocantos: `!s.Bool()`

marcelocantos: Might be worth changing `Bool` to `IsTrue`. Raise an issue.

nofun97: is the purpose of `Bool` to show the `Bool` value of a set or to show if it is not empty? If it's the latter, might be better to change it to `NotEmpty`.

nofun97:
Just remembered I'll have to use the function at the str lib

rel/value_set.go
marcelocantos: All values have a "truthiness" to them, which determines how they affect logical operations. `0`, `()` and `{}` are all considered "false". Every other value is considered "true". The `Bool()` method expresses this, but `IsTrue` would be less confusing.

PR #48, opened 2020-03-04
change the syntax to allow import from root

tiagralee:
Fixes # .

- change the syntax to allow import from root

Checklist:
- [x] Added related tests


ChloePlanet:
Current implementation doesn't indeed support import file from root. 
i.e. 
```bash
tree .
.
├── add.arrai
├── go.mod
└── pkg
    └── simple.arrai
```
It panics with `open add.arrai: no such file or directory` when you run `arrai e "///add"` under `./pkg`.

syntax/import.go
```diff
@@ -38,6 +52,29 @@ func importLocalFile() rel.Value {
 	return importLocalFileVar
 }
 
+func findProjectRootDir(currentDir string) (string, error) {
```
anz-gordonj7: sysl does a very similar thing to find the .sysl directory, have a look at how it implements it @ https://github.com/anz-bank/sysl/blob/master/pkg/loader/loader.go#L105 .

Using the string functions and assuming "/" for paths will break on windows. Also, https://golangcode.com/check-if-a-file-exists/ is a bit of a better way to check for file existing, again this code will break on windows which allows case insensitive filename.

syntax/import.go
```diff
@@ -38,6 +55,33 @@ func importLocalFile() rel.Value {
 	return importLocalFileVar
 }
 
+func findRootFromModule(modulePath string, fs afero.Fs) (string, error) {
```
anz-gordonj7: no need to use the afero.Fs here, use the os/file/path api directly please

marcelocantos: A big part of the reason for introducing afero was testability. Is that not applicable here?

tiagralee: @marcelocantos make sense, afero can avoid file I/O in unit tests. The latest version is using native path api, do you want me to change it back? Otherwise, I'll merge this pull request if it looks good to you.

PR #54, opened 2020-03-05
add basic os and file stdlib

nofun97:
- Add basic `os` and `file` functions
    - `cwd`
    - `file`
    - `pathSeparator`
    - `pathListSeparator`
    - `args`
    - `getenv`

### DIFF
--- a/syntax/std_os.go
+++ b/syntax/std_os.go
@@ -9,8 +9,8 @@ import (
 func stdOs() rel.Attr {
 	return rel.NewAttr("os", rel.NewTuple(
 		rel.NewAttr("args", getArgs()),
-		rel.NewAttr("pathSeparator", pathSeparator()),
-		rel.NewAttr("pathListSeparator", pathListSeparator()),
+		rel.NewAttr("path_separator", pathSeparator()),
+		rel.NewAttr("path_list_separator", pathListSeparator()),
 		rel.NewAttr("cwd", cwd()),
 		rel.NewNativeFunctionAttr("file", func(value rel.Value) rel.Value {
 			f, err := ioutil.ReadFile(value.(rel.String).String())
@@ -19,6 +19,6 @@ func stdOs() rel.Attr {
 			}
 			return rel.NewBytes(f)
 		}),
-		rel.NewNativeFunctionAttr("getenv", getEnv),
+		rel.NewNativeFunctionAttr("get_env", getEnv),
 	))
 }
--- a/syntax/std_os.go
+++ b/syntax/std_os.go
@@ -1,0 +1,24 @@
+package syntax
+
+import (
+	"io/ioutil"
+
+	"github.com/arr-ai/arrai/rel"
+)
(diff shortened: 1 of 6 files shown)


syntax/std_os.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/arr-ai/arrai/rel"
+)
+
+func stdOs() rel.Attr {
+	return rel.NewAttr("os", rel.NewTuple(
+		rel.NewAttr("args", getArgs()),
+		rel.NewAttr("pathSeparator", pathSeparator()),
+		rel.NewAttr("pathListSeparator", pathListSeparator()),
+		rel.NewAttr("cwd", cwd()),
+		rel.NewNativeFunctionAttr("file", func(value rel.Value) rel.Value {
```
anz-gordonj7: how will this be used? maybe it shold be named `readfile` ?

nofun97: `//.os.file("path/to/file")` which will then return an array of bytes.

@marcelocantos wanted a more declarative name instead of an action.

nofun97: documentation will be added

anz-gordonj7: My initial assumption would be that os.file() would return an object, not the file contents. what if you want to write to that file?

nofun97: that's true, might need to rethink the name. Regarding about writing to a file, I had this discussion with @marcelocantos. He would like to keep the language separated from the outside world and writing to a file is certainly going to be implemented but it will be more like a subcommand in the arrai cli instead of as part of the language.

marcelocantos: Arr.ai is a stateless language. The name was chosen deliberately to convey the notion that the expression `//.os.file("foo")` represents the contents of file `foo`, not the action of reading it. In a future version, we may even capture the idea of the file changing over time, with data flow analysis triggering updates to expressions that depend on the contents.

marcelocantos: Wrt writing to files, arr.ai has no concept of changing the outside world. It is an expression language that takes inputs and evaluates an expression in relation to them. The arrai program currently embodies this idea by not offering any mechanism to create/edit/delete files, talk to databases, send messages, push to rest services, etc., etc. For the most part, it can only return a result on stdout. In fact, even on the codegen front, arr.ai cannot output multiple files directly. Instead, you pass a map to `//.archive.tar.tar` http://bit.ly/2vIWzYu.

This is not the target for arr.ai, and I do plan to think about how arr.ai can be more flexible in terms of interacting with the world, but it needs to in line with arr.ai's core design as an expression language.

syntax/std_os_wasm.go
```diff
@@ -0,0 +1,27 @@
+// +build wasm
```
marcelocantos: Could we weave `ctx` throughout the API and pass an `afero.Fs` wrapping a JavaScript object? Raise an issue.

Issue #41, opened 2020-02-27
Leverage Go's standard library to flesh out the arr.ai standard library.


marcelocantos:
# In scope

* [ ] **bytes:**            implements functions for the manipulation of byte slices.
* [ ] **compress/bzip2:**   implements bzip2 decompression.
* [ ] **compress/gzip:**    implements reading and writing of gzip format compressed files, as specified in RFC 1952.
* [ ] **encoding/base64:**  implements base64 encoding as specified by RFC 4648.
* [ ] **encoding/csv:**     reads and writes comma-separated values (CSV) files.
* [ ] **encoding/json:**    implements encoding and decoding of JSON as defined in RFC 7159.
* [ ] **encoding/xml:**     implements a simple XML 1.0 parser that understands XML name spaces.
* [ ] **flag:**             implements command-line flag parsing.
* [ ] **path:**             implements utility routines for manipulating slash-separated paths.
* [ ] **regexp:**           implements regular expression search.
* [ ] **strconv:**          implements conversions to and from string representations of basic data types.
* [ ] **strings:**          implements simple functions to manipulate UTF-8 encoded strings.
* [ ] **tar:**              implements access to tar archives.
* [ ] **time:**             provides functionality for measuring and displaying time.
* [ ] **zip:**              provides support for reading and writing ZIP archives.

# Investigate

* [ ] **builtin:**              provides documentation for Go's predeclared identifiers.
* [ ] **compress/flate:**       implements the DEFLATE compressed data format, described in RFC 1951.
* [ ] **compress/lzw:**         implements the Lempel-Ziv-Welch compressed data format, described in T. A. Welch, “A Technique for High-Performance Data Compression”, Computer, 17(6) (June 1984), pp 8-19.
* [ ] **compress/zlib:**        implements reading and writing of zlib format compressed data, as specified in RFC 1950.
* [ ] **crypto:**               collects common cryptographic constants.
* [ ] **crypto/aes:**           implements AES encryption (formerly Rijndael), as defined in U.S. Federal Information Processing Standards Publication 197.
* [ ] **crypto/cipher:**        implements standard block cipher modes that can be wrapped around low-level block cipher implementations.
* [ ] **crypto/des:**           implements the Data Encryption Standard (DES) and the Triple Data Encryption Algorithm (TDEA) as defined in U.S. Federal Information Processing Standards Publication 46-3.
* [ ] **crypto/dsa:**           implements the Digital Signature Algorithm, as defined in FIPS 186-3.
[...]

PR #57, opened 2020-03-06
strconv lib

nofun97:
- add `strconv` library `eval` and `unsafe_eval`

Currently `eval` will do the exact same thing `unsafe_eval` does. A validator needs to be implemented in `eval` so that it can tell if it is given a complex expression.

Checklist:
- [x] Added related tests


syntax/std_strconv.go
```diff
@@ -0,0 +1,22 @@
+package syntax
+
+import (
+	"github.com/arr-ai/arrai/rel"
+)
+
+func stdStrconv() rel.Attr {
+	return rel.NewAttr("strconv", rel.NewTuple(
+		//TODO: eval needs to be changed to only evaluate simple expression
```
anz-gordonj7: 1) Is there a use-case for splitting this functionality into two funcs? Generally the stdlib should be as simple as possible.
2) what does `unsafe` in this context mean? 
3) What is expected to happen if an "unsafe" string is passed to `eval()`?

nofun97: 1 and 2. So the idea is to not give too much power to `eval`. `eval` should only be able to evaluate `Value`.  `unsafe_eval` is for the user when they need more power to evaluate expressions, but there are many problems to thing as expressions can vary wildly and it might lead to lots of problems (the example that was given is the halting problem), so it gives the idea to the user that they can use it to evaluate expression but it's up to them. As for how simple stdlib should be, there's an issue open #41 for this, might be better to discuss there.
3. As of right now, `eval()` will just evaluates it and returns a value. But, a validator will need to be implemented in the future. Right now the idea is to parse it and walk the AST and check whether it's a value or a complex expression.

syntax/std_strconv.go
```diff
@@ -0,0 +1,22 @@
...
+		// e.g. no functions, no math operations etc only simple values
+		rel.NewNativeFunctionAttr("eval", unsafeEval),
+		rel.NewNativeFunctionAttr("unsafe_eval", unsafeEval),
```
marcelocantos: The signature for `unsafe_eval` should be `\scope \expr ...`, where `scope` is a tuple of bound names to values to be used as the scope for `expr`.

syntax/std_strconv.go
marcelocantos: `eval` and `unsafe_eval` are roughly analogous to `eval` and `JSON.parse`, respectively, in JavaScript.

marcelocantos: I'm now thinking that `unsafe_eval` isn't that unsafe. JavaScript `eval` can do untold damage, including hack browser security, leak or delete user data, etc. However, because arr.ai is a hermetic expression language, `unsafe_eval` can't do anything harmful. Perhaps it makes more sense to call them `eval_value` and `eval_expr`. And yes they need to be different. See my other comment.

syntax/std_strconv.go
```diff
@@ -0,0 +1,22 @@
+package syntax
+
+import (
+	"github.com/arr-ai/arrai/rel"
+)
+
+func stdStrconv() rel.Attr {
+	return rel.NewAttr("strconv", rel.NewTuple(
```
marcelocantos: It doesn't feel quite right to have these in `strconv`. I've suggested elsewhere. that `eval` should be `eval.value` and `unsafe_eval` should be `eval.expr`. I think it therefore makes sense for `eval` to be the top level package.

nofun97:
`//.eval.expr` is deemed out of scope for now.

Issue #60, opened 2020-03-08
Make use of afero.Fs in context so that main build is safe for wasm


nofun97:
### Current arrai Version
There are two builds, `wasm` and non `wasm` build. The `wasm` build is meant to be used for web version of arrai.

### Use-cases
For arrai to only have one build that is safe for all use including web.

marcelocantos:
Yes. No need to rework any inflight PRs for this.

PR #61, opened 2020-03-08
wasm safe //.os.file

nofun97:
Fixes issue mentioned from the [comment](https://github.com/arr-ai/arrai/issues/60#issuecomment-596186873)  .

- Add wasm safe build for `//.os.file`

### DIFF
--- a/syntax/std_os_wasm.go
+++ b/syntax/std_os_wasm.go
@@ -25,3 +25,7 @@ func pathListSeparator() rel.Value {
 func cwd() rel.Value {
 	return rel.NewSet()
 }
+
+func file(rel.Value) rel.Value {
+	return rel.NewBytes([]byte{})
+}
--- a/syntax/std_os_wasm.go
+++ b/syntax/std_os_wasm.go
@@ -7,25 +7,25 @@ import (
 )
 
 func getArgs() rel.Value {
-	return rel.NewSet()
+	panic("not implemented")
 }
 
 func getEnv(value rel.Value) rel.Value {
-	return rel.NewSet()
+	panic("not implemented")
 }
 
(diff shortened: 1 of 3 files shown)


syntax/std_os_wasm.go
```diff
@@ -25,3 +25,7 @@ func pathListSeparator() rel.Value {
 func cwd() rel.Value {
 	return rel.NewSet()
 }
+
+func file(rel.Value) rel.Value {
+	return rel.NewBytes([]byte{})
```
marcelocantos: `panic("not implemented")`

Consider doing the same for the existing functions in this file. `cwd` definitely should panic. While you're at it, replace `rel.NewSet()` with `rel.None`.

syntax/std_os_wasm.go
```diff
@@ -7,25 +7,25 @@ import (
 )
 
 func getArgs() rel.Value {
-	return rel.NewSet()
+	panic("not implemented")
```
marcelocantos:
```go
return rel.None
```

syntax/std_os_wasm.go
```diff
@@ -7,25 +7,25 @@ import (
...
 }
 
 func getEnv(value rel.Value) rel.Value {
-	return rel.NewSet()
+	panic("not implemented")
```
marcelocantos:
```go
return rel.None
```

syntax/std_os_wasm.go
marcelocantos: Actually, nvm. It's fine as is.

syntax/std_os_wasm.go
marcelocantos: Actually, nvm. It's fine as is.

PR #63, opened 2020-03-10
fix get operation with quoted field names

nofun97:
Fixes the bug where `get` operations would panic if you use a field surrounded by quotes (e.g. `(x: 42).'x'`.

- Handled `STR` in get operations
- Corresponding tests

Checklist:
- [x] Added related tests

TODO:

Handle `(x: 42).*`. Not entirely sure what getting `*` would do but it is allowed in the grammar.


nofun97:
Fixed in #73

PR #69, opened 2020-03-19
#66 Add cache for local file import.

### LINKED ISSUE #66: add cache to avoid multiple identified imports
### Current arrai Version
v0.1.0 4add794

### Use-cases
Add cache to avoid multiple identified imports. e.g.
```
...
///foo/bar.math.pi
...
///foo/bar.math.pi
...
```
only import `///foo/bar.math` once

ericzhang6222:
Fixes #66  .

- Add cache for import behavior
- 

Checklist:
- [x] Added related tests

### DIFF
--- a/syntax/import.go
+++ b/syntax/import.go
@@ -152,7 +152,7 @@ func importURL(url string) (rel.Value, error) {
 			return val, nil
 		}
 		// Pass space as file path as it is remote url and can't be found in local env
-		val := bytesValue("", data)
+		val := bytesValue(NoPath, data)
 		cache.add(url, val)
 		return val, nil
 	}
@@ -173,7 +173,7 @@ func fileValue(filename string) (rel.Value, error) {
 
 func bytesValue(filename string, data []byte) rel.Value {
 	// maybe filename is "", so add check first
-	if len(strings.TrimSpace(filename)) > 0 {
+	if filename != NoPath {
 		if exists, val := cache.exists(filename); exists {
 			return val
 		}
@@ -189,7 +189,7 @@ func bytesValue(filename string, data []byte) rel.Value {
 	}
 
 	// maybe filename is "", so add check first
-	if len(strings.TrimSpace(filename)) > 0 {
+	if filename != NoPath {
 		cache.add(filename, value)
 	}
 	return value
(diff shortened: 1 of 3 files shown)


syntax/import.go
```diff
@@ -147,8 +148,13 @@ func importURL(url string) (rel.Value, error) {
 		if err != nil {
 			return nil, err
 		}
-
-		return bytesValue("", data), nil
+		if exists, val := cache.exists(url); exists {
+			return val, nil
+		}
+		// Pass space as file path as it is remote url and can't be found in local env
+		val := bytesValue("", data)
```
ChloePlanet: Sorry I just read the master branch code and realized we should use `NoPath` here instead of empty string.

ChloePlanet:
```suggestion
		val := bytesValue(NoPath, data)
```

syntax/import.go
```diff
@@ -166,6 +172,13 @@ func fileValue(filename string) (rel.Value, error) {
 }
 
 func bytesValue(filename string, data []byte) rel.Value {
+	// maybe filename is "", so add check first
+	if len(strings.TrimSpace(filename)) > 0 {
```
ChloePlanet:
```suggestion
	if filename != NoPath {
```

syntax/import.go
```diff
@@ -175,5 +188,9 @@ func bytesValue(filename string, data []byte) rel.Value {
 		panic(err)
 	}
 
+	// maybe filename is "", so add check first
+	if len(strings.TrimSpace(filename)) > 0 {
```
ChloePlanet:
```suggestion
	if filename != NoPath {
```

syntax/import.go
ChloePlanet: Same as the first comment.

syntax/import.go
ericzhang6222: Updated.

syntax/import.go
ericzhang6222: Fixed.

syntax/import.go
ericzhang6222: Fixed.

PR #71, opened 2020-03-25
Canonicalise tuple and set implementation types

### LINKED ISSUE #70: Canonicalise internal types
**Do not post any internal, closed source snippets** on this public issue tracker!

### Use-case

Arr.ai supports special syntax for certain constructs that aren't direct types in the language. This issue is concerned with three such cases: strings, arrays and dictionaries (dicts). These constructs are aliases for relations with a specific structure:

- `"abc"` = `{|@,@char| (0, 97), (1, 98), (2, 99)}`
- `[11, 12, 13]` = `{|@,@item| (0, 11), (1, 12), (2, 13)}`
- `{"a": 42, "b": 56}` = `{|@,@value| ("a", 42), ("b", 56)}`

Dealing with these aliases internally is quite complex and inefficient, because, for instance, code wanting to deal with strings has to detect whether a given type is either a `rel.String` or simply a `rel.genericSet` that happens to conform to the structure of a string.

### Resolution

Rework the internal type machinery such that, whenever a set that conforms to the structure of a string emerges, it is automatically canonicalised to the `rel.String` type. Likewise for sets conforming to the array and dict structures.

Currently, `rel.String` cannot hold every possible set comprising exclusively `(@, @char)` tuples. E.g., `{|@,@char| (0, 97), (0, 98)}` is unsupported. Since this is not strictly a string, it's OK for now to ignore this problem. It can be addressed later with intelligent management of overlapping characters. Same goes for `rel.Array` with duplicate indices. OTOH, `rel.Dict` supports duplicate keys now.

marcelocantos:
Fixes #70.

> **Note for reviewers:** This is a whale of a change and I have little hope of reworking the commit history in any reasonable time frame. The history is somewhat structured, though, so you'll probably still find it easier to review each commit rather than the entire change in one go.

- Implement custom tuple types for `(@: ..., @char: ...)`, `(@: ..., @item: ...)` and `(@: ..., @value: ...)`, corresponding to string characters, array items and dict values, respectively.
- Ensure that whenever a tuple is created that conforms to one of the above structures, it is always returned as the corresponding custom tuple type.
- Implement custom set types for strings, arrays and dicts.
- Ensure that whenever a set is created that contains exclusively tuples of one of the above custom types, it is always returned as the corresponding custom set type.

Checklist:
- [x] Added related tests
- [x] Made corresponding changes to the documentation (none required)


rel/expr_string_char_entry.go
```diff
@@ -0,0 +1,38 @@
...
+func (e StringCharTupleExpr) Eval(local Scope) (Value, error) {
+	at, err := e.at.Eval(local)
+	if err != nil {
+		return nil, err
+	}
+	char, err := e.char.Eval(local)
+	if err != nil {
+		return nil, err
+	}
+	return NewStringCharTuple(int(at.(Number).Float64()), rune(char.(Number).Float64())), nil
```
nofun97: why does `at` needs to be a number in a tuple?

rel/json.go
```diff
@@ -116,59 +116,20 @@ func jsonEscape(value Expr) interface{} {
...
+			return false
+		}
+		if x.Equal(True) {
+			return true
+		}
+		array := make([]interface{}, x.Count())
+		i := 0
+		for e := x.Enumerator(); e.MoveNext(); {
+			array[i] = jsonEscape(e.Current())
+			i++
```
nofun97: this might be cleaner
```go
array := make([]interface{}, 0, x.Count())
for e := x.Enumerator(); e.MoveNext(); {
    array = append(array, jsonEscape(e.Current())
}
```

rel/value_set_array.go
```diff
@@ -1,40 +1,321 @@
...
+
+// Kind returns a number that is unique for each major kind of Value.
+func (a Array) Kind() int {
+	return arrayKind
+}
+
+// Bool returns true iff the tuple has attributes.
+func (a Array) IsTrue() bool {
+	if len(a.values) == 0 {
+		panic("Empty string not allowed (should be == None)")
```
nofun97: is this supposed to be `Empty array ...`?

rel/value_set_array.go
```diff
@@ -1,40 +1,321 @@
...
+			return true
+		}
+		if bv.Less(av) {
+			return false
+		}
+	}
+	return len(a.values) < len(b.values)
+}
+
+// Negate returns {@neg: b}.
```
nofun97: is this supposed to be `{@neg: a}`?

rel/value_set_array.go
```diff
@@ -1,40 +1,321 @@
...
+// was already absent, the original Array is returned.
+func (a Array) Without(value Value) Set {
+	if t, ok := value.(ArrayItemTuple); ok {
+		if i := a.index(t.at); i >= 0 && t.item == a.values[i] {
+			if t.at == a.offset {
+				return Array{values: a.values[1:], offset: a.offset + 1}
+			}
+			if t.at == a.offset+i {
+				return Array{values: a.values[:i], offset: a.offset}
+			}
```
nofun97: this seems like if you're doing a `Without` with values in the index `i`, it would just get all the values until `i` and discard the rest. Is that the idea?

rel/value_set_array.go
```diff
@@ -1,40 +1,321 @@
...
+	return a
+}
+
+// Map maps values per f.
+func (a Array) Map(f func(v Value) Value) Set {
+	result := NewSet()
+	for e := a.Enumerator(); e.MoveNext(); {
+		result = result.With(f(e.Current()))
+	}
+	return result
```
nofun97: nit: `result` should be a `[]Value` and the function returns `NewSet(result...)`

rel/value_set_array.go
```diff
@@ -1,40 +1,321 @@
...
-// 			"@":           MatchInt(func(i int) { index = i }),
-// 			ArrayItemAttr: Bind(&item),
-// 		},
-// 		Lit(EmptyTuple),
-// 	).Match(v)
-// 	return
-// }
+func (e *arrayEnumerator) Offset() int {
+	return int(e.t.MustGet("@").(Number).Float64())
+}
```
nofun97: wouldn't `Index()` make more sense?

rel/value_tuple_array_item_test.go
```diff
@@ -0,0 +1,263 @@
...
+		},
+		{name: "no-@",
+			want: ArrayItemTuple{},
+			ok:   false,
+			args: args{t: NewTuple(NewIntAttr("at", 0), NewAttr(ArrayItemAttr, None))},
+		},
+		{name: "no-ArrayItemAttr",
+			want: ArrayItemTuple{},
+			ok:   false,
+			args: args{t: NewTuple(NewIntAttr("@", 0), NewIntAttr("item", 0))},
```
nofun97: nit: add newline before `name`

rel/value_tuple_dict_entry.go
```diff
@@ -0,0 +1,211 @@
...
+// the given name.
+func (t DictEntryTuple) Without(name string) Tuple {
+	if name == "@" || name == DictValueAttr {
+		return t.asGenericTuple().Without(name)
+	}
+	return t
+}
+
+func (t DictEntryTuple) Map(f func(Value) Value) Tuple {
+	return maybeNewDictEntryTupleFromTuple(t.asGenericTuple().Map(f))
```
nofun97: why use the `maybe...`?

rel/value_tuple_str_char_test.go
```diff
@@ -0,0 +1,270 @@
...
+		{name: "no-@",
+			want: StringCharTuple{},
+			ok:   false,
+			args: args{t: NewTuple(NewIntAttr("at", 0), NewIntAttr(StringCharAttr, 'a'))},
+		},
+		{name: "no-CharAttr",
+			want: StringCharTuple{},
+			ok:   false,
+			args: args{t: NewTuple(NewIntAttr("@", 0), NewIntAttr("char", 'a'))},
+		},
```
nofun97: newline before `name`

syntax/expr_order_test.go
```diff
@@ -4,7 +4,7 @@ import "testing"
 
 func TestOrder(t *testing.T) {
 	AssertCodesEvalToSameValue(t, `[-3, -2, -1, 0, 1, 2, 3]`, `{3, 2, 1, 0, -1, -2, -3} order \a \b  a < b`)
-	AssertCodesEvalToSameValue(t, `[3, 2, 1, 0, -1, -2, -3]`, `{3, 2, 1, 0, -1, -2, -3} order \a \b  a > b`)
+	// AssertCodesEvalToSameValue(t, `[3, 2, 1, 0, -1, -2, -3]`, `{3, 2, 1, 0, -1, -2, -3} order \a \b  a > b`)
```
nofun97: remove comment if not necessary?

syntax/expr_tuple_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"testing"
+
+	"github.com/arr-ai/arrai/rel"
+)
+
+func TestTupleType(t *testing.T) {
+	t.Parallel()
+	AssertCodeEvalsToType(t, rel.StringCharTuple{}, `(@: 1, @char: 65)`)
+	AssertCodeEvalsToType(t, rel.ArrayItemTuple{}, `(@: 1, @item: 2)`)
+	AssertCodeEvalsToType(t, rel.DictEntryTuple{}, `(@: {1, 2}, @value: 2)`)
```
nofun97: add failure test

rel/expr_array.go
```diff
@@ -0,0 +1,57 @@
...
+func (e ArrayExpr) Elements() []Expr {
+	panic("unfinished")
+}
+
+// String returns a string representation of the expression.
+func (e ArrayExpr) String() string {
+	var b bytes.Buffer
+	b.WriteByte('[')
+	i := 0
+	for _, expr := range e.elements {
```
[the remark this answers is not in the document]
marcelocantos: Wow that's embarrassing. Must have been copy pasted from my earliest days as a Go programmer. :-)

rel/expr_array_item_entry.go
```diff
@@ -0,0 +1,38 @@
+package rel
+
+import "fmt"
+
+// ArrayItemTupleExpr represents a single name:expr in a dictEntryTupleExpr.
+type ArrayItemTupleExpr struct {
+	at   Expr
+	item Expr
```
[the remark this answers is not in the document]
marcelocantos: The comment is misleading. I've updated it.

rel/expr_string_char_entry.go
marcelocantos: It represents the index of the character within a string.

rel/json.go
marcelocantos: Agreed. Probably more code from my Go infancy.

rel/value_set_array.go
marcelocantos: If removal is from either end, we can continue using `Array`. If not, we have to revert to a `genericSet`.

rel/value_set_array.go
marcelocantos: Map is defined over the elements of the set (that is, `(@: ..., @item: ...)`), not the items of the array.

nofun97: I meant something like this
```go
        result := make([]Value)
	for e := a.Enumerator(); e.MoveNext(); {
		result = append(result, f(e.Current()))
	}
	return NewSet(result...)
```

but after looking at `NewSet` implementation, apparently it's the same thing.

rel/value_set_array.go
nofun97: I can't see how `t.at == a.offset+i` means it's removing the item at the end of the array. It seems to me `i` can be in an index in them middle of the array and if it is in the middle, the function returns `Array{values: a.values[:i], offset: a.offset}`, what about `a.values[i+1:]`?

and also, why revert to a `genericSet`? It makes more sense (to me) to keep using `Array`. If I'm removing a value at index `i`, I'm expecting to get `Array{values: append(a.values[:i], a.values[i+1:]), offset: a.offset}`, also in this case, for each item in `a.values[i+1:]`, the index needs to be subtracted by one.

marcelocantos: Well spotted! PTAL

rel/value_tuple_dict_entry.go
marcelocantos: Again, very well spotted! This was probably a copy-paste mistake. For strings and arrays, it's possible for `Map` to produce a non-conforming tuple. This can't happen with `DictEntryTuple`.

syntax/expr_order_test.go
marcelocantos: Accidentally left it commented out.

syntax/expr_tuple_test.go
marcelocantos: Damn, you're on a roll.

PR #77, opened 2020-03-29
Eval output

### LINKED ISSUE #75: Output values as valid arr.ai syntax.

### Description

Currently:

```bash
$ arrai e '"abc"'
abc
```
which seems fine. However:

```
$ arrai e '[42] ++ ["abc"]'
[42, abc]
```
which doesn't seem fine.

The output looks like the syntax for an array, and it is in fact valid syntax, but it doesn't mean the same as the original. If this output is incautiously copy-pasted into a context in which the name `abc` just happens to be in scope, it could lead to a silent bug that is difficult to detect and diagnose.

#### Special case string printing

The correct solution isn't super obvious. If the output stage always produces correct syntax, it becomes impossible to output bare text to stdout. My current inclination is to treat strings as a special case when printing, as per Python:

```bash
$ python -c 'print("abc")'
abc
$ python -c 'print([42, "abc"])'
[42, 'abc']
```

To avoid edge case hassles, the empty set would have to be deemed an empty string for printing purposes.

#### Command line option

A command line option could control the output approach. It could default to arr.ai syntax with an option to output strings bare, or vice versa.

#### Always text, require explicit formatting

Only support printing strings as bare text. Other values would first have to be converted to a string via something like `$"${value}"`.

marcelocantos:
Strings (including the empty set) print bare.
Everything else prints as a valid arr.ai value.

Fixes #75.

- Print strings as bare text. Treat empty set as empty string.
- Print everything else as a valid arr.ai value.
- Fix tuple expr canonicalization.

Checklist:
- [x] Added related tests
- [x] Made corresponding changes to the documentation no

### DIFF
--- a/rel/expr_tuple.go
+++ b/rel/expr_tuple.go
@@ -159,7 +159,8 @@ func (e *TupleExpr) Eval(local Scope) (Value, error) {
 			return nil, err
 		}
 	}
-	return tuple, nil
+	// TODO: Construct new tuple directly
+	return tuple.(*GenericTuple).Canonical(), nil
 }
 
 // Get returns the Expr for the given name or nil if not found.
--- a/rel/value_repr.go
+++ b/rel/value_repr.go
@@ -159,7 +159,7 @@ func reprValue(v Value, w io.Writer) {
 	case Number:
 		reprNumber(v, w)
 	default:
-		panic(fmt.Errorf("Repr(): unexpected Value type %T: %[1]v", v))
+		panic(fmt.Errorf("Repr(): unexpected Value type %T: %[1]v", v)) //nolint:golint
 	}
 }
 
--- a/rel/value_repr.go
+++ b/rel/value_repr.go
@@ -6,12 +6,9 @@ import (
 	"strings"
 )
 
-var reprEscapes = map[byte][]byte{
-	'\'': escapesWithDelim('\''),
-	'"':  escapesWithDelim('"'),
-}
+var reprEscapes = escapesWithDelim()
 
-func escapesWithDelim(delim byte) []byte {
+func escapesWithDelim() []byte {
 	escapes := make([]byte, 32)
 	escapes['\a'] = 'a'
 	escapes['\b'] = 'b'
@@ -41,7 +38,6 @@ func reprOffset(offset int, w io.Writer) {
 }
 
 func reprEscape(s string, delim byte, w io.Writer) {
-	escapes := reprEscapes[delim]
 	fmt.Fprintf(w, "%c", delim)
 	// TODO: optimise by handling non-special characters in groups.
 	for _, c := range s {
@@ -49,7 +45,7 @@ func reprEscape(s string, delim byte, w io.Writer) {
 			fmt.Fprintf(w, `\%c`, c)
 		} else if c >= 32 {
 			fmt.Fprintf(w, "%c", c)
-		} else if escape := escapes[c]; escape != 0 {
+		} else if escape := reprEscapes[c]; escape != 0 {
 			fmt.Fprintf(w, `\%c`, escape)
 		} else {
 			fmt.Fprintf(w, `\x%02x`, c)
--- a/rel/value_repr.go
+++ b/rel/value_repr.go
@@ -69,6 +69,11 @@ func reprString(str String, w io.Writer) {
 	}
 }
 
+func reprBytes(b Bytes, w io.Writer) {
+	reprOffset(b.offset, w)
+	fmt.Fprint(w, b.String())
+}
+
 func reprArray(a Array, w io.Writer) {
 	reprOffset(a.offset, w)
 	fmt.Fprint(w, "[")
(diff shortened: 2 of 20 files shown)


rel/expr_tuple.go
```diff
@@ -159,7 +159,8 @@ func (e *TupleExpr) Eval(local Scope) (Value, error) {
 			return nil, err
 		}
 	}
-	return tuple, nil
+	// TODO: Construct new tuple directly
+	return tuple.(*GenericTuple).Canonical(), nil
```
anzboi: As far as I can tell, the only utility of `Canonical` is to potentially convert the tuple into a list-item/string-char/etc tuple.

rel/value_repr.go
```diff
@@ -0,0 +1,170 @@
...
+	case ArrayItemTuple:
+		reprArrayItemTuple(v, w)
+	case DictEntryTuple:
+		reprDictEntryTuple(v, w)
+	case *GenericTuple:
+		reprTuple(v, w)
+	case Number:
+		reprNumber(v, w)
+	default:
+		panic(fmt.Errorf("unknown Value type %T: %[1]v", v))
```
anzboi: More helpful error message?
```suggestion
		panic(fmt.Errorf("repr does not support value type %T: %[1]v", v))
```

rel/value_repr.go
```diff
@@ -0,0 +1,170 @@
...
+	escapes['\n'] = 'n'
+	escapes['\r'] = 'r'
+	escapes['\t'] = 't'
+	escapes['\v'] = 'v'
+	return escapes
+}
+
+type reprCommaSep bool
+
+func (s *reprCommaSep) Sep(w io.Writer) {
```
anzboi: Thats pretty nifty

rel/value_repr_test.go
```diff
@@ -0,0 +1,24 @@
+package rel
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRepr(t *testing.T) {
```
anzboi: More tests? seems pretty light considering how many types just got a repr function.
Most importantly, it seems tuples are completely missing (sets, strings, arrays, dicts, and some nested elements thereof are tested in one way or another).

rel/value_tuple.go
```diff
@@ -122,6 +122,15 @@ func NewXML(tag []rune, attrs []Attr, children ...Value) Tuple {
 	return EmptyTuple.With("@xml", b.Finish())
 }
 
+func (t *GenericTuple) Canonical() Tuple {
```
anzboi: Would it be worth making this part of the `Tuple` interface
pro: can be called on any tuple without knowing if it is specific or not.
con: basically a noop for anything thats not a GenericTuple.

IMO its not worth it, the type cast and `ok` check is good enough. Just putting the idea out there.

rel/value_tuple.go
```diff
@@ -157,30 +166,26 @@ var LexerNamePat = `([$@A-Za-z_][0-9$@A-Za-z_]*)`
 
 var identRE = regexp.MustCompile(`\A` + LexerNamePat + `\z`)
 
+func TupleNameRepr(name string) string {
+	if identRE.Match([]byte(name)) {
+		return name
+	}
+	data, err := json.Marshal(name)
```
anzboi: Is the purpose of `json.Marshal` to put quotes around it and escape some special characters?

syntax/test_helpers.go
```diff
@@ -53,7 +53,7 @@ func AssertCodeEvalsToType(t *testing.T, expected interface{}, code string) bool
 	}
 	codeExpr := pc.CompileExpr(ast)
```
anzboi: I cannot comment at the top of this function.

This test utility functions name doesn't match the comment.
> // AssertCodeEvalsToType asserts that the exprs evaluate to the same value.
Should read
> // AssertCodeEvalsToType asserts that the code evaluates to the expected type.
I checked the call to `AssertExprEvalsToType`, it does indeed only check the type

rel/value_repr.go
```diff
@@ -0,0 +1,174 @@
...
+	"io"
+	"strings"
+)
+
+var reprEscapes = map[byte][]byte{
+	'\'': escapesWithDelim('\''),
+	'"':  escapesWithDelim('"'),
+}
+
+func escapesWithDelim(delim byte) []byte {
```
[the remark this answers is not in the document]
anzboi: Already resolved 👍

rel/expr_tuple.go
marcelocantos: That's exactly what it's for. Ideally, such forms should be detected upfront during tuple construction, which is happening already (`Canonical()` cheats by simply invoking it.) but sometimes you have an existing tuple that becomes a canonicalizable form through some transform and this method makes it easy to accommodate these scenarios.

Eventually, we need to tune this so that all tuple operations automatically return the canonical type and this function becomes redundant. But we're not there yet.

rel/value_repr.go
marcelocantos: Good point. I went for a slightly different error message.

rel/value_tuple.go
marcelocantos: Would be worth considering but for the fact that it's temporary hack.

rel/value_tuple.go
marcelocantos: Existing code that was refactored. The better approach is to use repr, which I've just done.

PR #78, opened 2020-03-29
Misc test improvements

marcelocantos:


syntax/test_helpers.go
```diff
@@ -59,19 +59,16 @@ func AssertCodeEvalsToType(t *testing.T, expected interface{}, code string) bool
 	return true
 }
 
-// AssertCodePanics asserts that the code panics when executed.
+// AssertCodePanics asserts that the exprs evaluate to the same value.
```
dancantos: I think the old comment was more accurate

syntax/test_helpers.go
```diff
@@ -59,19 +59,16 @@ func AssertCodeEvalsToType(t *testing.T, expected interface{}, code string) bool
...
-		return false
-	}
-	return true
+	return assert.Panics(t, func() {
+		pc := ParseContext{SourceDir: ".."}
+		ast, err := pc.Parse(parser.NewScanner(code))
+		if assert.NoError(t, err, "parsing code: %s", code) {
+			codeExpr := pc.CompileExpr(ast)
+			codeExpr.Eval(rel.EmptyScope) //nolint:errcheck
+		}
```
nofun97: you can use `require.NoError` instead.

nofun97: to replace `assert.NoError`

marcelocantos: That fails the entire test. I want to allow the test to continue after this assertion fails.

PR #79, opened 2020-03-29
Fix string enum bug

marcelocantos:
- Fix a bug enumerating strings with offsets

Checklist:
- [x] Added related tests
- [x] Made corresponding changes to the documentation (none required)

### DIFF
--- a/rel/value_set_generic.go
+++ b/rel/value_set_generic.go
@@ -70,6 +70,17 @@ func NewSet(values ...Value) Set {
 	return set
 }
 
+func CanonicalSet(s Set) Set {
+	if s, ok := s.(GenericSet); ok {
+		values := make([]Value, 0, s.Count())
+		for e := s.Enumerator(); e.MoveNext(); {
+			values = append(values, e.Current())
+		}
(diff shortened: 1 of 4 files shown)


rel/value_set_generic.go
```diff
@@ -70,6 +70,17 @@ func NewSet(values ...Value) Set {
...
+	if s, ok := s.(GenericSet); ok {
+		values := make([]Value, 0, s.Count())
+		for e := s.Enumerator(); e.MoveNext(); {
+			values = append(values, e.Current())
+		}
+		return NewSet(values...)
+	}
+	return s
+}
+
```
nofun97: I don't get how this fixes the enumerating offset string bug, is this a different feature?

marcelocantos: It's related. The fix is in the enumerator logic.

nofun97: yep I saw the fix (I'm assuming it's the `Current` function), but how is `CanonicalSet` related? It looks to me it's just taking value from a `GenericSet` and make another `Set`.

PR #82, opened 2020-03-29
JSON to arrai object

nofun97:
JSON Mapping

Maps JSON values to arrai objects
- `42` -> `42`
- `true/false` -> `(b: true/false)`
- `null` -> `(null: {})` 
- `"abc"` -> `(s: "abc")`
- `[1, 2, 3]` -> `(a: [1, 2, 3])`
- `{"a": 42}` -> `{"a": 42}`

Checklist:
- [ ] Added related tests

### DIFF
--- a/syntax/std_json_test.go
+++ b/syntax/std_json_test.go
@@ -1,0 +1,42 @@
+package syntax
+
+import "testing"
+
+func TestJSONDecode(t *testing.T) {
+	t.Parallel()
+	AssertCodePanics(t, `//.json.decode(123)`)
+	AssertCodesEvalToSameValue(t,
+		`{
+			"a": (s: "string"),
+			"b": 123,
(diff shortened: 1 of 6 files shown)


syntax/std_json_test.go
```diff
@@ -0,0 +1,42 @@
...
+			"c": 123.321,
+			"d": (a: [1, (s: "string again"), (a: []), {}]),
+			"e": {
+				"f": {
+					"g": (s: "321")
+				},
+				"h": (a: [])
+			},
+			"i": (null: {}),
+			"j": (a: [(b: {()}), (b: {})]),
```
marcelocantos: You can write `true` instead of `{()}` and `false` instead of `{}`. They are exactly the same thing, but it conveys the intent better.

PR #87, opened 2020-03-31
Terminate let with required ";" instead of optional "in".

marcelocantos:
It's more concise and gives a more natural feel to the code, especially after making it a required change. (Jsonnet offers a precedent.)

Checklist:
- [x] Added/updated related tests


marcelocantos:
> There are several `let` statements in `intro.md` needs to be updated, like L381, L389, L400, L404, etc.

That was careless of me! Hopefully I've got them all now.

PR #94, opened 2020-04-02
Parallelise tests requiring remote connectivity

marcelocantos:
Speeds up testing by parallelising tests that connect to github.

### DIFF
--- a/syntax/import_cache.go
+++ b/syntax/import_cache.go
@@ -1,28 +1,64 @@
 package syntax
 
-import "github.com/arr-ai/arrai/rel"
+import (
+	"sync"
+
+	"github.com/arr-ai/arrai/rel"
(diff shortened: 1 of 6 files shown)


syntax/import_cache.go
```diff
@@ -1,28 +1,68 @@
...
+	for {
+		if val, has := service.cache[key]; has {
+			if val != nil {
+				return val
+			}
+			// Another goroutine is adding an entry.
+			service.cond.Wait()
+		} else {
+			break
+		}
```
ChloePlanet, line 48:
```suggestion
		}
		
		break
```

marcelocantos: We need to continue after `service.cond.Wait()`, not break.

PR #95, opened 2020-04-02
Implement //.os.stdin

marcelocantos:
When accessed, `//.os.stdin` reads stdin into a byte array.

Checklist:
- [x] Added related tests

### DIFF
--- a/rel/value_tuple_str_char.go
+++ b/rel/value_tuple_str_char.go
@@ -1,4 +1,4 @@
-package rel
+package rel //nolint:dupl
 
 import (
 	"fmt"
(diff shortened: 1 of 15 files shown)


rel/value_tuple_str_char_test.go
```diff
@@ -1,4 +1,4 @@
-package rel
+package rel //nolint:dupl
```
nofun97, line 1: same with this

rel/value_tuple_str_char.go
```diff
@@ -1,4 +1,4 @@
-package rel
+package rel //nolint:dupl
```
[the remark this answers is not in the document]
marcelocantos: No, several blocks of code in the file are.

rel/value_tuple_str_char_test.go
marcelocantos: ditto

PR #120, opened 2020-04-14
Tidy up import path syntax

marcelocantos:
- Reinterpret `//./a.b` as equivalent to `(//./a).b` (currently equivalent to`//./'a.b'`.
- Consequently, importing files with extensions will require string-name syntax: `//./'a.b'`.

Checklist:
- [x] Added/updated related tests

### DIFF
--- a/syntax/package_test.go
+++ b/syntax/package_test.go
@@ -41,11 +41,11 @@ func TestPackageImportFromRoot(t *testing.T) {
 }
 
 func TestJsonPackageImportFromModuleRoot(t *testing.T) {
-	AssertCodesEvalToSameValue(t, `{'location': 'Melbourne', 'name': 'foo'}`, `///examples/json/foo.json`)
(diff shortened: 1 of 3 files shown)


syntax/package_test.go
```diff
@@ -41,11 +41,11 @@ func TestPackageImportFromRoot(t *testing.T) {
 }
 
 func TestJsonPackageImportFromModuleRoot(t *testing.T) {
-	AssertCodesEvalToSameValue(t, `{'location': 'Melbourne', 'name': 'foo'}`, `///examples/json/foo.json`)
+	AssertCodesEvalToSameValue(t, `{'location': 'Melbourne', 'name': 'foo'}`, `///examples/json/'foo.json'`)
 }
```
tiagralee, line 45: The syntax much clean now, so we will only gonna support file with extension wrapped with quote?

PR #139, opened 2020-04-18
Includes source context in arr.ai evaluation error

### LINKED ISSUE #136: Includes source context when arr.ai evaluations crash
### Current arrai Version
v0.1.0

### Use-cases
Includes filename, line number, column number in evaluation crash error message.

### References
- #88

ChloePlanet:
Fixes #136  .

- Add Scanner function to `type Expr interface`
- Embed struct `ExprScanner` which implements `Scanner` function into Exprs and Values.
- Add source context including filename, line number and column number to `Eval` error

before:
image

after:
image


cmd/arrai/eval.go
```diff
@@ -19,7 +19,7 @@ var evalCommand = &cli.Command{
 }
 
 func evalImpl(source string, w io.Writer) error {
-	return evalExpr(syntax.NoPath, source, w)
+	return evalExpr(".", source, w)
```
marcelocantos, line 22: Is this change necessary? The original intent was to disallow imports when there is no source file.

ChloePlanet: Yes. Actually `"."` is the original code. I thought `syntax.NoPath` would be better so I changed it to NoPath. Then I found out `arrai e "//./examples/simple/simple"` cannot find the right file path with `NoPath`.

marcelocantos: Sorry, I meant to retract this yesterday when I realised it makes perfect sense, but forgot. All good.

PR #150, opened 2020-04-20
Revise conditional operator

### LINKED ISSUE #114: Revise conditional operator syntax

### Use-cases
Current conditional operator syntax inverts the reading flow, placing the conditional after the successful case: `x if cond1 else y if cond2 else z`. This confuses even readers who are well versed in this syntactic form.

Change the syntax to a form that follows the natural reading order.

Proposed syntax: `cond (cond1: x, cond2: y, *: z)`. Also consider supporting a control expression: `control_var cond (1: x, 2: y, *: z)`. In both cases `*: z` is optional, defaulting to `*: false` (which is the same as {}).

ericzhang6222:
Fixes #114  .

- Add new conditional operator `cond`, `cond (cond1: x, cond2: y, *: z)` and `*:z` is optional, `condVar cond (1: x, 2: y, *: z)` is not implemented in this PR, 
- Add waring to deprecate operator if else, `x if cond1 else y if cond2 else z`. 

Checklist:
- [x] Added related tests


rel/expr_cond.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	b.WriteByte('}')
+	return b.String()
+}
+
+// Eval returns the true condition. It must have only one true condition.
+func (e *CondExpr) Eval(local Scope) (Value, error) {
+	var trueCond *DictEntryTupleExpr
+	// Evaluates the valid condition, only one condition whose isTrue() == true can be valid.
+	// If there is not any valid condition, the condtion defaultExpr will work.
```
marcelocantos, line 52: Multiple true conditions are valid, even multiple instances of the literal `true`. This can happen when code-generating arr.ai code. The `cond` expression returns the value for the first condition that evaluates to `true`, going from left to right. A linter could be written to complain about such cases.

I agreed that there must be at most one `*` condition and there must be no other conditions after it.

rel/expr_if_else.go
```diff
@@ -5,11 +5,13 @@ import (
 )
 
 // IfElseExpr returns the tuple applied to a function.
+// Deprecated: IfElseExpr will be removed, it should use CondExpr instead.
 type IfElseExpr struct {
 	ifTrue, cond, ifFalse Expr
 }
 
 // NewIfElseExpr returns a new IfElseExpr.
+// Deprecated: NewIfElseExpr will be removed, it should use NewCondExpr instead.
```
marcelocantos, line 14: Emit a warning the first time this is called.

syntax/compile.go
```diff
@@ -217,6 +222,10 @@ func (pc ParseContext) compileRbinop(b ast.Branch, c ast.Children) rel.Expr {
 }
 
 func (pc ParseContext) compileIf(b ast.Branch, c ast.Children) rel.Expr {
+	log.Error(context.Background(),
```
marcelocantos, line 225: It's not an error. Use `log.Print`

marcelocantos, line 225: `sync.Once`

rel/expr_cond.go
```diff
@@ -0,0 +1,89 @@
...
+	dicExpr, defaultExpr Expr
+}
+
+// NewCondExpr returns a new CondExpr.
+func NewCondExpr(dict Expr, defaultExpr Expr) Expr {
+	return &CondExpr{dict, defaultExpr}
+}
+
+// String returns a string representation of the expression.
+func (e *CondExpr) String() string {
```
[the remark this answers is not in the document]
ericzhang6222: Fixed.

rel/expr_if_else.go
marcelocantos: nvm, it's logged elsewhere.

syntax/compile.go
ericzhang6222: I think Error is better, as it can get user's attention and consider this issue seriously. I do like to use Warning, but there is not Warning API in log.

rel/expr_cond.go
ericzhang6222: Fixed.

syntax/compile.go
ericzhang6222: The expected code is:
```
	loggingOnce.Do(func() {
		log.Error(context.Background(), "...")
	})
```
right?

PR #155, opened 2020-04-21
Fix #128, add the slice feature to all types of Set

### LINKED ISSUE #128: Function-call slice syntax

### Use-cases

Python's `[]` operator supports slicing: `a[start:stop:step]`.

Add the same capability to arr.ai. Since arr.ai treats arrays as functions, `a(index)`, slicing support must be added to function-call syntax: `a(start:stop:step)`.

nofun97:
Fixes #128 .

- Add Slicing features to Dictionary, Set, and Array

Checklist:
- [x] Added related tests


rel/value_set_generic.go
```diff
@@ -335,6 +336,49 @@ func (s GenericSet) Call(arg Value) Value {
...
+			return array.CallSlice(start, end, step, inclusive)
+		case dictEntryTupleType:
+			tuples := make([]DictEntryTuple, 0, s.set.Count())
+			for i := s.Enumerator(); i.MoveNext(); {
+				tuples = append(tuples, i.Current().(DictEntryTuple))
+			}
+			return NewDict(true, tuples...).CallSlice(start, end, step, inclusive)
+		}
+	}
+	panic(fmt.Sprintf("Set %s is not indexed", s))
```
nofun97, line 379: should this function panic or just return `None`?

docs/slicing.md
```diff
@@ -0,0 +1,143 @@
...
+## Usages
+
+Only certain types of sets can be used with the slicing expression. These sets are:
+
+1. `Array`
+2. `String`
+3. `Byte Array`
+4. `Dictionary`
+
+Generic sets (i.e. unindexed sets) or any other expression that does not evaluates to the
```
cuminandpaprika, line 26:
```suggestion
Generic sets (i.e. unindexed sets) or any other expression that does not evaluate to the
```

cuminandpaprika: Also, can you give an example of an unindexed set here?

docs/slicing.md
```diff
@@ -0,0 +1,143 @@
...
+
+The slicing expression can be used by putting the expression as an argument to a call.
+
+The syntax of the slicing expression is as follows:
+
+```text
+start=expr? ";" end=expr? (";" step=expr)?
+```
+
+As shown above, you can define a slice by defining a lower bound, an upper bound, both or none.
```
cuminandpaprika, line 13: Maybe some details about how the lower bound and upper bound are both inclusive?

docs/slicing.md
```diff
@@ -0,0 +1,143 @@
+# Slicing
+
+Most sets in arrai can be sliced with the slice expression.
+
+The slicing expression can be used by putting the expression as an argument to a call.
+
+The syntax of the slicing expression is as follows:
+
+```text
+start=expr? ";" end=expr? (";" step=expr)?
```
cuminandpaprika, line 10: I would probably replace this with

```someSet(lowerIndex;upperIndex;optionalStepsize)```

File: docs/slicing.md [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+```
+
+As shown above, you can define a slice by defining a lower bound, an upper bound, both or none.
+You can also define a `step` that is used to define the increment of the numbers from the lower
```
cuminandpaprika, line 14:
```suggestion
You can also define a `stepsize` that is used to define the increment of the numbers from the lower
```

docs/slicing.md
nofun97: I'll put it in the behaviour section

docs/slicing.md
nofun97: changed to 
```
someSet(optionalLowerBoundExpr;optionalUpperBoundExpr;optionalStepSizeExpr)
```

docs/slicing.md
```diff
@@ -0,0 +1,143 @@
...
+
+When `start` or `end` are defined, the values will be used. However, when the range
+is invalid (i.e `start > end && step > 0` or `start < end && step < 0`), an empty
+set will be returned as the result of the slice expression.
+
+When `start` or `end` are not defined, the value of `step` will determine their
+values.
+
+If `step` is positive, `start` will have the lowest possible value. In
+`Array` and `String`, the value of `start` defaults to the offset value. In
```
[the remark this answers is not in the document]
nofun97: it basically offsets the indexes, this might give a better explanation https://github.com/arr-ai/arrai/blob/master/docs/intro.md#some-lexical-conventions

docs/slicing.md
```diff
@@ -0,0 +1,143 @@
...
+The syntax of the slicing expression is as follows:
+
+```text
+start=expr? ";" end=expr? (";" step=expr)?
+```
+
+As shown above, you can define a slice by defining a lower bound, an upper bound, both or none.
+You can also define a `step` that is used to define the increment of the numbers from the lower
+bound to the upper bound. When `step` is not defined, its value defaults to `1`.
+
```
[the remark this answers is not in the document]
nofun97: there's a section for this for the `Invalid Index`

docs/slicing.md
```diff
@@ -0,0 +1,143 @@
...
+### Behaviour
+
+#### Default values
+
+Slice expression behaves differently when provided with different expressions.
+
+When the lower bound, the upper bound, and step are provided, the values will have
+to evaluate to a number as slicing can only be done with number expressions.
+Anything other than `Number`, the expression will fail.
```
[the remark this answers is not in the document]
nofun97: currently there's no error message yet. Will add it along with the doc changes.

docs/slicing.md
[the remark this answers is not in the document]
cuminandpaprika: Ah, I see, maybe a link to that in the docs could help?

docs/slicing.md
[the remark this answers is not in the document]
cuminandpaprika: Yup, so I like the fact that at the top you start general, without going into too many details. But maybe a touch more detail about what valid lower and upper bounds might be would help them understand they need to dig in further below

docs/slicing.md
```diff
@@ -0,0 +1,151 @@
...
+{1, 2, 3, "abc"}
+```
+
+Since the expression is also an argument, the following expression is also valid
+
+```text
+[1, 2, 3, 4, 5](;;-1, 0)
+```
+
+this expression reverses the array and fetch the first value. In
```
cuminandpaprika, line 58:
```suggestion
this expression reverses the array and fetches the first value. In
```
Oops one last thing I missed

syntax/parse.go
```diff
@@ -75,6 +72,7 @@ expr   -> C* amp="&"* @ C* arrow=(
         | C* STR C*
         | C* NUM C*;
 nest   -> C* "nest" names IDENT C*;
+range  -> start=expr? ";" end=expr? (";" step=expr)?;
```
marcelocantos, line 75: `slice` seems a better name.

docs/slicing.md
```diff
@@ -0,0 +1,151 @@
...
+
+```text
+someSet(optionalLowerBoundExpr;optionalUpperBoundExpr;optionalStepSizeExpr)
+```
+
+As shown above, you can define a slice by defining a lower bound, an upper bound, both or none.
+You can also define a `stepSize` that is used to define the increment of the numbers from the lower
+bound to the upper bound. `stepSize` defaults to `1` when it is not defined.
+
+Lower bounds may be negative and upper bounds may be larger than the set size. In such cases, the index will wraparound. Refer to the example below.
```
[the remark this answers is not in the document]
nofun97: so slicing `[:-1]` shouldn't be allowed?

nofun97:
The behaviour of the slice is updated, please take a look on the documentation for slicing. The new behaviour removes wraparound and also added some changes in dictionary slicing. A test utility also needed to be changed due to the new source context changes, a proper fix is required by checking if the error message is contained in the actual error is currently enough.

docs/slicing.md
```diff
@@ -0,0 +1,145 @@
...
+bound to the upper bound. `stepSize` defaults to `1` when it is not defined.
+
+## Usages
+
+Only certain types of sets can be sliced. These sets are:
+
+1. `Array`
+2. `String`
+3. `Byte Array`
+4. `Dictionary`
```
marcelocantos, line 20: Any binary relation with an `@` attribute should support function call syntax. Raise an issue and link here if this isn't the case.

docs/slicing.md
```diff
@@ -0,0 +1,145 @@
...
+someSet(optionalLowerBoundExpr;optionalUpperBoundExpr;optionalStepSizeExpr)
+```
+
+As shown above, you can define a slice by defining a lower bound, an upper bound, both or none.
+You can also define a `stepSize` that is used to define the increment of the numbers from the lower
+bound to the upper bound. `stepSize` defaults to `1` when it is not defined.
+
+## Usages
+
+Only certain types of sets can be sliced. These sets are:
```
marcelocantos, line 15: Functions should be slice-able too. If not, raise an issue and link here.

docs/slicing.md
```diff
@@ -0,0 +1,145 @@
...
+#### Default values
+
+Slice operation behaves differently when provided with different expressions.
+
+Lower bound is always inclusive while upper bound is always exclusive. As in it
+always includes values that corresponds to ranges of values from `start` to
+`end - 1`. This is true for all scenarios, whether `start` or `end` are defined or not.
+However, when `end` is not provided, the last element is always included. This is not
+a special case as default value of `end` will include the last element. Refer to the
+explanations below.
```
marcelocantos, line 65:
```suggestion
When `end` is omitted, the slice operation consumes every element,
`[start, start + step, start + 2*step, ...], till it reaches a value that doesn't
have a corresponding element.
```

marcelocantos, line 65: If this mirrors current behaviour, I'm happy to document it as is. However, `end` should be a special case, since step might take on odd values, like `0.01`, which would break the described semantic. If necessary, raise an issue to rectify this and link it here.

docs/slicing.md
```diff
@@ -0,0 +1,145 @@
+# Slicing
+
+Functions (which includes strings, arrays, etc.) can be called using slice syntax, The syntax is as follows:
+
+```text
+someSet(optionalLowerBoundExpr;optionalUpperBoundExpr;optionalStepSizeExpr)
```
marcelocantos, line 6: Referring to upperBound, lowerBound and stepSize everywhere is overly verbose. Just use `start`, `stop` and `step` everywhere.

docs/slicing.md
```diff
@@ -0,0 +1,145 @@
...
+
+Lower bound is always inclusive while upper bound is always exclusive. As in it
+always includes values that corresponds to ranges of values from `start` to
+`end - 1`. This is true for all scenarios, whether `start` or `end` are defined or not.
+However, when `end` is not provided, the last element is always included. This is not
+a special case as default value of `end` will include the last element. Refer to the
+explanations below.
+
+When the lower bound, the upper bound, and step are provided, the values will have
+to evaluate to a number as slicing can only be done with number expressions.
```
marcelocantos, line 68:
```suggestion
`start`, `stop` and `step` must be numbers.
```

docs/slicing.md
nofun97: I'm currently limiting it to those data types and any `GenericSet` that only contain binary relation with an `@` attribute. Other than that, the expression will fail, I'll include it in the docs.

nofun97: It currently only supports sets that can be converted to `Array`, `String`, `Byte Array`, or `Dictionary`. Anything other than that, the expression will fail. Should I still raise an issue?

docs/slicing.md
nofun97: current implementation doesn't take into account for odd values of `end`. Will raise an issue

PR #159, opened 2020-04-25
Basic interactive shell

marcelocantos:
- Interactive shell for evaluating expressions
- Basic tab-completion for standard library (`//.<tab>`)

Checklist:
- [too hard] Added related tests


anzboi:
Question: Does the shell allow for persistence of identifiers define through `let`. Reason being is I want to define a lambda and try it out on some various input. I don't want to have to re-type the function every time.

Other features that might help this
* `up` copies to previously evaluated input
* referencing previously evaluate values (I could evaluate a lambda definition and reference via some syntax).

PR #190, opened 2020-05-04
Pattern matching

### LINKED ISSUE #132: Pattern-matching name binding

### Use-cases

Also known as de-structuring. Currently, `let a = expr` binds the name `a` to the value of `expr`. An extension on this is to unpack a complex value into multiple names within a pattern. For instance, `let [a, b, c] = expr` performs two distinct functions:

1. Checks that `expr` evaluates to an array with three elements.
2. Binds a, b and c to elements 0, 1 and 2 of the array, respectively.

### Scope of applicability

Pattern matching can be applied wherever names can be bound:

- Let binding: `let pattern = ...`
- Function parameters: `\pattern ...`

Additionally, case expressions may be extended to support name-binding: `case (pattern1: ..., pattern2: ...)`. Patterns are tried in left-to-right order, choosing the first successful match.

### Pattern syntax

- [x] match single number in let statements #190
- [x] `"hello"`, `1 + 2`:  match the value of the given expression. #396
- [x] `{a, 42}` matches a set with two elements, one of which is 42 and binds the name `a` to the other element. #252
- [x] `{(a), (b), (c)}` matches a set with three elements and binds names `a`, `b` and `c` to the three elements. Which element binds to which name is non-deterministic. #252
- [x] `{(a), (b), (c), ...t}` as above, but allows sets with > 3 elements, binding the remainder to `t`. Which three elements are bound to which names may be non-deterministic. #252
- [x] `[a, b, c]` matches an array with three items and binds names `a`, `b` and `c` to items 0, 1 and 2, respectively. #190
- [x] `[a, b, c, ...t]` allows extra elements in addition to the explicitly bounds ones and binds `t` to an array of them. #238 
- [x] `[a, ...t, b, c]`. The `...t` part can appear anywhere. #238 
- <del>`[a, b, ...t, c, d, ...u, e, f]`. Multiple `...` parts may lead to non-deterministic bindings, depending on the constraints imposed by patterns, a&ndash;f.</del>
- [x] `(m: x, n: y)` matches a tuple with attributes `m` and `n` and binds name `x` to the `m` attribute and name `y` to the `n` attribute. #236 #256 #287 
- [x] `(m: x, n: y, ...t)` allows extra elements in addition to the explicitly bounds ones and binds `t` to a tuple of them. #238 
- [x] `{a: x, b: y}` matches a dict with keys `a` and `b` and binds name `x` to the `a` key and name `y` to the `b` key. Unlike for tuples, this pattern allows `a` and `b` to also be patterns, which may lead to non-deterministic bindings. #237 
- [x] `{a: x, b: y, ...t}` allows extra elements in addition to the explicitly bounds ones and binds `t` to a dict of them. #238 
- <del>`a <: S` matches any value in set `S` and binds the name `a` to it.</del>
   - `S` may be a bounded finite set or an unbounded set representing a type, such as `//number`.
- [x] Recursively, `a`, `b`, `c` `t`, `u`, `x` and `y`, above, can be patterns.

In all cases, `..._` (allow a "tail" but don't bind it) may be shortened to `...`.

Non-deterministic forms should be left unimplemented for now. We need to think carefully about the implications of non-deterministic name-binding. A first pass at supporting these forms might only bind possibly-non-deterministic forms if the actual binding turns out to be deterministic wrt the given value. For example:

1. `let {x, y} = v` is non-deterministic no matter what `v` is. This should be disallowed. 
1. `let {(id: 0, ...t), (id: 1, ...u)} = ...` is always deterministic and therefore always allowable, since all matching values of `v` will bind `a` and `b` deterministically.
1. `let {[x, y], [z, ...]} = v` is deterministic if `v` is `{[1, 2], [3]}` but non-deterministic if `v` is `{[1, 2], [3, 4]}`.

This suggests that support for non-deterministic forms can be implemented in the following stages:

1. Disallow outright.
1. Allow forms that are guaranteed to be deterministic, given their sub-patterns.
1. Allow all forms, but fail at runtime if the given value leads to a non-deterministic binding.

#### Variable controls

It may be necessary to use existing bound names to control pattern matching. For instance, instead of `{a, 42}`, perhaps the controlling value needs to be a variable instead of the literal 42: `let control = 42; let {a, control} = ...`. In this case, how do we indicate that `a` is a new name to be bound whereas `control` is a value to be matched against? The solution is to ensure that it is an expression rather than a name. The simplest way to achieve this is to wrap it in parentheses: `let {a, (control)} = ...`.

#### What about strings?

Pattern matching on portions of strings and byte arrays is trickier. Expression-string syntax would cover the most common uses:

```
let $"C-${custid}" = "C-1234"
```

Regexp syntax would be more powerful, but trickier to implement. The following captures the idea, but probably wouldn't work in practice, since `//re{...}` returns a tuple that won't match in the intended manner.

```
let //re{(?<lhs>\w+)\s*=\s*(?<rhs>\w+)} = "abc = def";
[lhs, rhs]
```

### A speculative but intriguing idea

An interesting way to achieve the above would be to support a custom pattern matching protocol. The basic idea is for `//re{...}` to return as a tuple with special attributes that trigger customer behaviour. The result would be equivalent to something like the following expression:

```text
let match = ...;
(
    source: `(?<lhs>\w+)\s*=\s*(?<rhs>\w+)`,
    :match,
    sub: ...,
    subf: ....
    @pattern: \value match(value).named,
)
```

The pattern matching system would evaluate the regexp and, on seeing that its output was a tuple containing a `@pattern` attribute, would pass the incoming value to that attribute. For the above example, the returned tuple would be `(lhs: "abc", rhs: "def")`. The pattern matching system would bind the names `lhs` and `rhs` to their respective values.

Naturally, library authors could leverage this to introduce pattern matching for just about any construct, e.g.: `let //time.date{y-m-d} = customer.dob`.

### LINKED ISSUE #235: pattern matching MVP
### Use-cases
The first step of https://github.com/arr-ai/arrai/issues/132.
Implement pattern matching with ident, number and array.

ChloePlanet:
Fixed #235 

- [x] match single number in let statements
```arrai
// correct - `42` matches the value 42 and doesn't bind any names.
let 42 = 42; 42
// wrong
let 42 = 3; 42
// wrong
let 42 = 3; 3
// wrong
let 42 = 42;
```
- [x] `[a, b, c]` matches an array with three items and binds names `a`, `b` and `c` to items 0, 1 and 2, respectively.
```
let [a, b, c] = [1, 2, 3]; [a, b, c](2)
let [a, b, c] = [1, 2, 3]; [c, b](0)
let [a, b, c] = [1, 2, 3]; b
```

(Maybe...)

- [ ] `[a, b, c, ...t]` allows extra elements in addition to the explicitly bounds ones and binds `t` to an array of them.
- [ ] `[a, ...t, b, c]`. The `...t` part can appear anywhere.

(Maybe not...)

- [ ] `[a, b, ...t, c, d, ...u, e, f]`. Multiple `...` parts may lead to non-deterministic bindings, depending on the constraints imposed by patterns, a&ndash;f.

Checklist:
- [ ] Added related tests


examples/simple/pattern.arrai
```diff
@@ -0,0 +1 @@
+let [a,(b)] = [1,2]; a
```
marcelocantos, line 1: Missing trailing \n

syntax/arrai.wbnf
```diff
@@ -68,3 +65,11 @@ NUM    -> /{ (?: \d+(?:\.\d*)? | \.\d+ ) (?: [Ee][-+]?\d+ )? };
...
 
 .wrapRE -> /{\s*()\s*};
+
+.macro patternterms(top) {
+    C* "{" C* rel=(names tuple=("(" v=top:",", ")"):",",?) "}" C*
+  | C* "{" C* set=(elt=top:",",?) "}" C*
+  | C* "{" C* dict=((key=top ":" value=top):",",?) "}" C*
+  | C* "[" C* array=(item=top:",",?) C* "]" C*
+  | C* "(" tuple=(pairs=(name? ":" v=top):",",?) ")" C*
+};
```
marcelocantos, line 75: Another one.

syntax/expr_let_test.go
```diff
@@ -2,9 +2,96 @@ package syntax
...
+		`let x = 3; let [(x)] = [3]; x`,
+	)
+	AssertCodesEvalToSameValue(t,
+		`2`,
+		`let x = 3; let [b, (x)] = [2, 3]; b`,
+	)
+	AssertCodesEvalToSameValue(t,
+		`2`,
+		`let x = 3; let [_, b, (x)] = [1, 2, 3]; b`,
+	)
```
marcelocantos, line 85: nit: These can all be one-liners.

PR #191, opened 2020-05-04
Fix #188, //rel.union accepts set of sets

### LINKED ISSUE #188: //rel.union needs to take generic set
### Current arrai Version
Currently `//rel.union` only takes ordered set, it needs to take in generic set.

### Use-cases
```
@> //rel.union({{1, 2}, {2, 3}, {3}})
{1, 2, 3}
```

nofun97:
Fixes #188  .

- `//rel.union` can now take an array of sets or set of sets

Checklist:
- [x] Added related tests

### DIFF
--- a/syntax/std_rel.go
+++ b/syntax/std_rel.go
@@ -9,7 +9,18 @@ func stdRel() rel.Attr {
 		rel.NewNativeFunctionAttr("union", func(v rel.Value) rel.Value {
 			s := v.(rel.Set)
 			sets := make([]rel.Set, 0, s.Count())
-			for e, ok := s.ArrayEnumerator(); ok && e.MoveNext(); {
+			var e rel.ValueEnumerator
+			switch u := v.(type) {
+			case rel.Array:
+				var ok bool
+				e, ok = u.ArrayEnumerator()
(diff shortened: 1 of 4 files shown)


syntax/std_rel.go
```diff
@@ -9,7 +9,18 @@ func stdRel() rel.Attr {
 		rel.NewNativeFunctionAttr("union", func(v rel.Value) rel.Value {
 			s := v.(rel.Set)
 			sets := make([]rel.Set, 0, s.Count())
-			for e, ok := s.ArrayEnumerator(); ok && e.MoveNext(); {
+			var e rel.ValueEnumerator
+			switch u := v.(type) {
+			case rel.Array:
```
marcelocantos, line 14: Only sets should be accepted.

nofun97: treating array as set created complications as `Enumerator.Current()` would return a tuple `(@, @values)`

marcelocantos: That's as intended.

nofun97: right, some of the example uses `//rel.union(array)`, I'll change it

PR #201, opened 2020-05-06
Fix #199, add command `arrai file.arrai`

### LINKED ISSUE #199: Allow implied run command

### Use-cases

```bash
$ arrai run foobar args
```

Runs foobar.arrai with args. Make it so that this does the same thing:

```bash
$ arrai foobar args
```

If the file happens to be called, e.g., `json`, always run the `json` command instead. If the user really wants to call their file `json` and still run it from arrai, they have two choices:

```bash
$ arrai run json
$ arrai ./json
```

It seems that this is already what happens when invoking via `#!...`.

```bash
$ cat json
#!/usr/bin/env arrai
//os.args
$ PATH=.:$PATH json
['./json']
```

nofun97:
Fixes #199 .

- you can run arrai script by doing the command `arrai path/to/file.arrai`

Checklist:
- [x] Added related tests

### DIFF
--- a/cmd/arrai/file_util.go
+++ b/cmd/arrai/file_util.go
@@ -8,9 +8,10 @@ import (
 
 func isExecCommand(arg string, cmds []*cli.Command) bool {
 	for _, cmd := range cmds {
-		names := append(make([]string, 0, 1+len(cmd.Aliases)), cmd.Name)
-		names = append(names, cmd.Aliases...)
-		for _, name := range names {
+		if arg == cmd.Name {
+			return true
+		}
+		for _, name := range cmd.Aliases {
 			if arg == name {
 				return true
 			}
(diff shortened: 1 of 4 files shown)


cmd/arrai/file_util.go
```diff
@@ -0,0 +1,29 @@
...
+
+func isExecCommand(arg string, cmds []*cli.Command) bool {
+	for _, cmd := range cmds {
+		names := append(make([]string, 0, 1+len(cmd.Aliases)), cmd.Name)
+		names = append(names, cmd.Aliases...)
+		for _, name := range names {
+			if arg == name {
+				return true
+			}
+		}
```
marcelocantos, line 17:
```suggestion
		if arg == cmd.Name {
			return true
		}
		for _, name := range cmd.Aliases {
			if arg == name {
				return true
			}
		}
```

PR #239, opened 2020-05-11
Fix closure binding

marcelocantos:
**Recommend reviewing commits individually.**

- cond with no match returns false
- Replace all if-else instances with cond
- Add missing scanners in expr constructors
- Fix closure binding

Checklist:
- [x] Added related tests


rel/pattern.go
```diff
@@ -44,7 +44,7 @@ func (p ValuePattern) Bind(scope Scope, value Value) Scope {
 		}
 	}
 
-	return scope.With(p.String(), value)
+	return EmptyScope
```
ChloePlanet, line 47:
```suggestion
	return EmptyScope.With(p.String(), value)
```

PR #291, opened 2020-05-20
add CallAll to Set interface

nofun97:
This PR adds `CallAll` to the interface `Set`

`CallAll(arg)` works on `Tuple`s inside `Set` which have the `@` attribute. It finds any tuple whose `@` attribute matches `arg` and return the corresponding values in a `Set`. If there are no corresponding values, it returns an empty `Set`.

Checklist:
- [ ] Added related tests

### DIFF
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -249,15 +249,6 @@ func (e *BinExpr) String() string {
 
 // Eval returns the subject
 func (e *BinExpr) Eval(local Scope) (_ Value, err error) {
-	// defer func() {
-	// 	switch r := recover().(type) {
-	// 	case nil:
-	// 	case error:
-	// 		panic(wrapContext(r, e))
-	// 	default:
-	// 		panic(wrapContext(fmt.Errorf("panic: %v", r), e))
-	// 	}
-	// }()
 	a, err := e.a.Eval(local)
 	if err != nil {
 		return nil, wrapContext(err, e)
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -153,11 +153,7 @@ func NewWhereExpr(scanner parser.Scanner, a, pred Expr) Expr {
 			if x, ok := a.(Set); ok {
 				if p, ok := pred.(Closure); ok {
 					return x.Where(func(v Value) bool {
-						match, err := p.Call(v, local)
-						if err != nil {
-							panic(err)
-						}
-						return match.IsTrue()
+						return SetCall(p, v).IsTrue()
 					}), nil
 				}
 				return nil, errors.Errorf("'where' rhs must be a Fn, not %T", a)
@@ -175,7 +171,7 @@ func NewOrderByExpr(scanner parser.Scanner, a, key Expr) Expr {
 				if k, ok := key.(Closure); ok {
 					values, err := OrderBy(x,
 						func(value Value) (Value, error) {
-							return k.Call(value, local)
+							return SetCall(k, value), nil
 						},
 						func(a, b Value) bool {
 							return a.Less(b)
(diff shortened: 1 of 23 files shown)


rel/value.go
```diff
@@ -108,12 +108,25 @@ type Set interface {
...
 	CallAll(arg Value) Set
 
 	ArrayEnumerator() (OffsetValueEnumerator, bool)
 }
 
+// SetCall does a CallAll to a Set and panics when there's less or more than 1 value.
+func SetCall(s Set, arg Value) Value {
+	result := s.CallAll(arg)
+	if !result.IsTrue() {
+		panic("no result")
```
anzdaddy, line 120: Make this panic message consistent with the next one.

rel/value.go
```diff
@@ -108,12 +108,25 @@ type Set interface {
...
+	}
+	if result.Count() > 1 {
```
anzdaddy, line 122: Avoid `Count` where possible, since it's O(&infin;) worst case. Enumerate up to two elements and fail if you reach the second one.

nofun97:
BinExpr used a Callable interface to evaluate a few expression
1. Function
2. Closure
3. ExprClosure
4. NativeFunction
5. NativeExprFunction

Everything except NativeExprFunction now implements Set which
uses SetCall to evaluate the resulting Value.

NativeExprFunction and anything that uses it are commented out.

rel/value_set_test.go
```diff
@@ -39,3 +25,36 @@ func TestAsString(t *testing.T) {
...
+		),
+		NewTuple(
+			NewAttr("@", NewNumber(4)),
+			NewAttr("random", NewNumber(5)),
+		),
+	)
+
+	AssertEqualValues(t, NewSet(NewNumber(42), NewNumber(24)), set.CallAll(NewNumber(1)))
+	AssertEqualValues(t, NewSet(NewNumber(3), NewNumber(4)), set.CallAll(NewNumber(2)))
+	AssertEqualValues(t, None, set.CallAll(NewNumber(3)))
```
marcelocantos, line 57: 2 and 3 should both fail. Only binary tuples with one `@` attribute should be supported.

rel/value_set_array_test.go
```diff
@@ -44,3 +44,29 @@ func TestArrayWithout(t *testing.T) {
...
+	AssertEqualValues(t, NewSet(NewNumber(12)), three.CallAll(NewNumber(0)))
+	AssertEqualValues(t, None, three.CallAll(NewNumber(1)))
+	AssertEqualValues(t, None, three.CallAll(NewNumber(-3)))
+
+	three = NewOffsetArray(2, NewNumber(10), NewNumber(11), NewNumber(12))
+	AssertEqualValues(t, NewSet(NewNumber(10)), three.CallAll(NewNumber(2)))
+	AssertEqualValues(t, NewSet(NewNumber(11)), three.CallAll(NewNumber(3)))
+	AssertEqualValues(t, NewSet(NewNumber(12)), three.CallAll(NewNumber(4)))
+	AssertEqualValues(t, None, three.CallAll(NewNumber(1)))
+	AssertEqualValues(t, None, three.CallAll(NewNumber(5)))
```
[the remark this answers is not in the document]
nofun97: arrays can't return multiple values

rel/expr_binary.go
```diff
@@ -290,15 +249,15 @@ func (e *BinExpr) String() string {
...
-	}()
+	// defer func() {
+	// 	switch r := recover().(type) {
+	// 	case nil:
+	// 	case error:
+	// 		panic(wrapContext(r, e))
+	// 	default:
+	// 		panic(wrapContext(fmt.Errorf("panic: %v", r), e))
+	// 	}
+	// }()
```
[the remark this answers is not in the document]
nofun97: forgot to remove that

rel/value_set_test.go
nofun97: they panic now

rel/value_set_array_test.go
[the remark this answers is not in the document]
marcelocantos: Ah, yes, sorry, confused it with dict.

PR #292, opened 2020-05-20
support dict pattern

### LINKED ISSUE #237: support dicts in pattern matching
### Use-cases
The third step of #132.
After tuples, dicts with fixed keys, e.g.: `let {"foo": f, "bar": b, ...} = t; [f, b]`

ChloePlanet:
Fixes #237.

- support dict pattern `let {"x": a, "y": b} = {"x": 4, "y": 5}; [a, b]`

Checklist:
- [x] Added related tests

### DIFF
--- a/rel/pattern.go
+++ b/rel/pattern.go
@@ -86,36 +86,31 @@ func (p ArrayPattern) String() string {
 	return b.String()
 }
 
-type AttrPattern struct {
+type TuplePatternAttr struct {
 	name    string
 	pattern Pattern
 }
 
-func NewAttrPattern(name string, pattern Pattern) AttrPattern {
-	return AttrPattern{
+func NewTuplePatternAttr(name string, pattern Pattern) TuplePatternAttr {
+	return TuplePatternAttr{
 		name:    name,
 		pattern: pattern,
 	}
 }
 
-func (p AttrPattern) Bind(scope Scope, value Value) Scope {
-	fmt.Println(value)
-	return scope
+func (a TuplePatternAttr) String() string {
+	return fmt.Sprintf("%s:%s", a.name, a.pattern)
 }
 
-func (p AttrPattern) String() string {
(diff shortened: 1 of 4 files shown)


rel/pattern.go
```diff
@@ -165,3 +160,66 @@ func (p TuplePattern) String() string {
 	b.WriteByte(')')
 	return b.String()
 }
+
+type DictPatternEntry struct {
```
anzdaddy, line 164: I was going to suggest that this is really a `DictEntryPattern`. But actually, it's both. :grin:

syntax/compile.go
```diff
@@ -176,14 +180,34 @@ func (pc ParseContext) compileTuplePattern(b ast.Branch) rel.Pattern {
...
+			if len(keyExprs) == len(valuePtns) {
+				entryPtns := make([]rel.DictPatternEntry, 0, len(keyExprs))
+				for i, keyExpr := range keyExprs {
+					entryPtns = append(entryPtns, rel.NewDictPatternEntry(keyExpr, valuePtns[i]))
+				}
+				return rel.NewDictPattern(entryPtns...)
+			}
+		}
+		panic("mismatch between dict keys and values")
+	}
```
anzdaddy, line 207: nit: could flatten this slightly:
```suggestion
	if (keys != nil) != (values != nil) {
		panic("mismatch between dict keys and values")
	}
	if (keys != nil) && (values != nil) {
		keyExprs := pc.compileExprs(keys.(ast.Many)...)
		valuePtns := pc.compilePatterns(values.(ast.Many)...)
		if len(keyExprs) == len(valuePtns) {
			entryPtns := make([]rel.DictPatternEntry, 0, len(keyExprs))
			for i, keyExpr := range keyExprs {
				entryPtns = append(entryPtns, rel.NewDictPatternEntry(keyExpr, valuePtns[i]))
			}
			return rel.NewDictPattern(entryPtns...)
	}
}
```

PR #313, opened 2020-05-24
Fix #285, add conditional accessor syntax

### LINKED ISSUE #285: Conditional accessor syntax

### Use-cases

Tuple accesses fail if the requested name is missing and function calls fail if the parameter is not in the function's domain. Working around these failures is awkward.

Define a syntax that resolves to a fallback value if a lookup failure occurs somewhere along a chain of accesses and function calls.

### Proposed Solution

Append each access in a chain with `?` and terminate it with `:` and a fallback value. E.g.: `a.b("c").d` will become `a.b?("c")?.d?:42`.

Notes:

1. `?` is not an operator that tests whether its LHS is some kind of "null" for false value. It is part of the access and function-call syntax. `x.y?:42` is valid syntax, but `x?:42` isn't.
1. There is no default fallback value. The chain must end in `?:<expr>`.

nofun97:
Fixes #285 .

- Adds conditional accessor syntax (look at #285 for more details)

Checklist:
- [ ] Added related tests

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -331,7 +331,7 @@ Arr.ai supports operations on numbers.
 
    Not all access failures are allowed. Only missing attributes of a `Tuple` or a `Set` call
    does not return exactly 1 value.
-   1. `(a: (b: 1)).a?.b?:42` will fail
+   1. `(a: (b: 1)).a?.b.c?:42` will fail as it will try to evaluate `1.c?:42`.
 5. Function slice:
    1. `[1, 1, 2, 3, 5, 8](2:5) = [2, 3, 5]`
    2. `[1, 2, 3, 4, 5, 6](1:5:2) = [2, 4]`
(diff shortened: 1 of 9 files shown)


docs/README.md
```diff
@@ -317,7 +317,22 @@ Arr.ai supports operations on numbers.
...
+   3. `{"a": 1}("b")?:42 = 42`
+   4. `{"a": 1}("a")?:42 = 1`
+
+   It also allows appending access expressions.
+   1. `(a: {"b": (c: 2)}).a?("b").c?:42 = 2`
+   2. `(a: {"b": (c: 2)}).a?("b").d?:42 = 42`
+
+   Not all access failures are allowed. Only missing attributes of a `Tuple` or a `Set` call
+   does not return exactly 1 value.
+   1. `(a: (b: 1)).a?.b?:42` will fail
```
marcelocantos, line 334: Explain why.

PR #319, opened 2020-05-25
fix #258, add syntax for character

### LINKED ISSUE #258: Character syntax
There is currently no way to express a Unicode code point as a character. In Erlang, `$A` is an alias for the number 65, which is the ASCII code for the letter `A`. We can't use:

| syntax | can't because... |
|:-:|:-|
| `'A'` | string |
| `$A` | valid name |
| `$'A'` | expression string |
| `${A}` | reserved for future use to allow non-c-ident names |
| `$:A` | confuses dict syntax `{$:A}` |

Current proposal: `%A`.

Other options:
| syntax | comment |
|:-:|:-|
| `@'A'` | Feels a bit heavy. |

nofun97:
Fixes #258 .

- Add syntax for character

Checklist:
- [x] Added related tests


docs/README.md
```diff
@@ -212,6 +212,22 @@ Arr.ai allows a shorthand form to represent relations:
...
+to the ASCII code of `char`.
+
+Usage:
+
+```arrai
+%a  = 97
+%A  = 65
+%\n = 10
+%\t = 9
+%🙂 = 240
```
marcelocantos, line 228: Nope

PR #332, opened 2020-05-26
Cond pattern matching

### LINKED ISSUE #288: pattern matching in cond
Several changes required here:

1. For consistency with `let` syntax, all conditions must have `(...)` syntax, not just multi-valued conditions:
    Before:
    ```
    let a = 1; let b = 2; x cond (a: 10, b: 20)
    ```
    After:
    ```
    let a = 1; let b = 2; x cond ((a): 10, (b): 20)
    ```
1. Remove `*` as a special case. `_` will work as a substitute without any special coding required.
1. Change the overall syntax from `cond (...)` and `expr cond (...)` to `cond {...}` and `expr cond {...}`. This to more clearly differentiate outer bracketing from the conditions: `x cond {(1): 10, (2): 20}`.

The `cond {...}` form without a LHS argument doesn't require pattern matching, since there is no value to match against. However, it's best to make the changes to both to maintain consistent syntax. 

Following piece will be done later:
Since `cond {...}` is simply a shorthand for `true cond {...}`, they should be given the same semantics, and so pattern matching should still work, e.g.: `cond {{t}: t}` should evaluate to `()`.

ericzhang6222:
Fixes #288  .

- 
- 

Checklist:
- [x] Added related tests


rel/expr_cond_pattern_control_var.go
```diff
@@ -0,0 +1,120 @@
...
+func (expr *CondPatternControlVarExpr) Eval(local Scope) (Value, error) {
+	varVal, err := expr.controlVarExpr.Eval(local)
+	if err != nil {
+		return nil, wrapContext(err, expr.controlVarExpr)
+	}
+
+	for index, condition := range expr.conditions {
+		switch condition := condition.(type) {
+		case Expr:
+			switch condition := condition.(type) {
```
marcelocantos, line 61: Flatten to a single `switch`.

syntax/arrai.wbnf
```diff
@@ -10,7 +10,7 @@ expr   -> C* amp="&"* @ C* arrow=(
         > C* @:binop="&&" C*
         > C* @:compare=/{!?(?:<:|=|<=?|>=?|\((?:<=?|>=?|<>=?)\))} C*
         > C* @ if=("if" t=expr ("else" f=expr)?)* C*
-        > C* @ cond=("cond" "(" (key=(("(" expr:",", ")") | expr) ":" value=expr):",",? ("*" ":" f=expr ","?)? ")")? C*
+        > C* @ cond=("cond" "{" (condition=(("(" expr:",", ")") | "("expr")" | pattern) ":" value=expr):",",? ("_" ":" f=expr ","?)? "}")? C*
```
marcelocantos, line 13: `"(" expr ")"` should be part of `patternterms`. `identpattern=IDENT` should be something like `exprpattern=expr`. We should tidy this up separately from this PR and then leverage it.

syntax/arrai.wbnf
```diff
@@ -20,7 +20,7 @@ expr   -> C* amp="&"* @ C* arrow=(
         > C* @ postfix=/{count|single}? C* touch? C*
         > C* (get | @) tail_op=(safe_tail | tail)* C*
         > %!patternterms(expr)
-        | C* cond=("cond" "(" (key=@ ":" value=@):",",? ("*" ":" f=expr ","?)? ")") C*
+        | C* cond=("cond" "{" (key="(" @ ")" ":" value=@):",",? ("_" ":" f=expr ","?)? "}") C*
```
marcelocantos, line 23: Since we're going to relax the rule that all expressions in patterns must be enclosed in `(...)`, let's remove the requirement here too.

syntax/compile.go
```diff
@@ -29,6 +29,8 @@ import (
 
 const NoPath = "\000"
 
+const ExprStr = "expr"
```
marcelocantos, line 32: Rename to `exprTag`.

rel/expr_cond_pattern_control_var.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/arr-ai/wbnf/parser"
+)
+
+// CondPatternControlVarExpr which is used for `cond` pattern matching.
+type CondPatternControlVarExpr struct {
+	ExprScanner
+	controlVarExpr Expr
+	conditions     []Pattern
+	values         []Expr
```
marcelocantos, line 15: Define a structure to hold condition/value pairs.

rel/pattern.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	"github.com/go-errors/errors"
 )
 
 // Pattern can be inside an Expr, Expr can be a Pattern.
 type Pattern interface {
 	// Require a String() method.
 	fmt.Stringer
 
-	Bind(scope Scope, value Value) Scope
+	Bind(scope Scope, value Value) (Scope, error)
```
marcelocantos, line 16: `ExprAsPattern` (below) should not be required anymore.

rel/pattern.go
```diff
@@ -322,3 +351,45 @@ func (p DictPattern) String() string {
 	b.WriteByte('}')
 	return b.String()
 }
+
+type ExprsPattern struct {
+	exprs []Expr
+}
+
+func NewExprsPattern(exprs []Expr) ExprsPattern {
```
marcelocantos, line 359:
```suggestion
func NewExprsPattern(exprs ...Expr) ExprsPattern {
```

syntax/test_helpers.go
```diff
@@ -121,3 +122,10 @@ func AssertScan(t *testing.T, l *Lexer, tok Token, intf interface{}, lexeme stri
 
 	return assert.Equal(t, lexeme, lexeme, l)
 }
+
+func AssertEvalExprString(t *testing.T, expected, source string) bool {
+	expr, err := Compile(".", source)
+	return assert.True(t, err == nil) &&
+		assert.True(t, expr != nil) &&
```
marcelocantos, line 129:
```suggestion
	return assert.NoError(t, err) &&
		assert.NotNil(t, expr) &&
```

examples/grpc/grpc.arrai
```diff
@@ -1,21 +1,21 @@
...
+        t.type cond {
+            ('primitive'): t.primitive cond {
+                ('DECIMAL'):  'double',
+                ('INT'):      'int64',
+                ('FLOAT'):    'double',
+                ('STRING'):   'string',
+                ('STRING_8'): 'string',
+                ('BOOL'):     'bool',
+                ('DATE'):     'string',
+                ('DATETIME'): 'string',
```
marcelocantos, line 12: I really don't want to force this on people. Let's add string and character literals alongside number literals as special cases not requiring parentheses. Expressions in general should still require them.

Issue #356, opened 2020-06-01
Inconsistent index out of range on holey arrays


orlade:
Please **do not post any internal, closed source snippets** on this public issue tracker!

### Description

Probably a duplicate but I'm not sure what of.

### Steps to Reproduce

```
@> [1,,3] >> \t t * 2
[2, , 6]
@> [1,,2] >> \t t * 2
2020-06-01T10:19:45.4082434+10:00 error_message=runtime error: index out of range [4] with length 4

.:1:8:
[1,,2] >> \t t * 2 INFO 
```

Also when mapping:

```
@> [1,3] >> \t t * 2
[2, 6]
@> [1,,3] >> \t t * 2
[2, , 6]
@> [1,,,3] >> \t t * 2
2020-06-01T10:31:19.9842297+10:00 error_message=runtime error: index out of range [-1]

.:1:9:
[1,,,3] >> \t t * 2 INFO
```

### Your Environments
```sh
$ go version
go version go1.14 linux/amd64

```

orlade-anz:
Urk, it's actually nondeterministic:

```
ladeo@Olivers-MacBook-Pro ~ % arrai eval '[1,,2] >> \t t * 2'
[2, , 4]
ladeo@Olivers-MacBook-Pro ~ % arrai eval '[1,,2] >> \t t * 2'
INFO[0000] runtime error: index out of range [4] with length 4

.:1:8:
[1,,2] >> \t t * 2
```

orlade-anz:
Turns out the non-determinism is related to enumerating the elements of a set when turning it into an array; depending on which element was picked first, it may or may not expose a bug in the conversion logic which doesn't account for the existence of holes.

PR #376, opened 2020-06-03
Add simple test for SequenceMapExp

orlade-anz:
WIP #356.

- Add a basic test for SequenceMapExpr which previously had none.

Checklist:
- [x] Added related tests

### DIFF
--- a/rel/expr_seqmap_test.go
+++ b/rel/expr_seqmap_test.go
@@ -6,23 +6,14 @@ import (
 	"github.com/arr-ai/wbnf/parser"
 )
 
-func TestSequenceMapExprIdent(t *testing.T) {
+func TestSequenceMapExprAccessors(t *testing.T) {
 	t.Parallel()
 
-	AssertExprsEvalToSameValue(t,
-		NewSequenceMapExpr(
-			*parser.NewScanner(""),
-			NewArray(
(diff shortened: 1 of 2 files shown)


nofun97:
so we test expressions in the `rel` package because we can just write the arrai code for the expression, this is for example https://github.com/arr-ai/arrai/blob/master/rel/expr_indexed_seqmap.go

orlade-anz:
Also WIP #381

rel/expr_seqmap_test.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"testing"
+
+	"github.com/arr-ai/wbnf/parser"
+)
+
+func TestSequenceMapExprAccessors(t *testing.T) {
+	t.Parallel()
+
+	lhs := NewArray(NewNumber(float64(1)))
```
nofun97, line 12:
```suggestion
	lhs := NewArray(NewNumber(1))
```

rel/expr_seqmap_test.go
```diff
@@ -0,0 +1,19 @@
...
+	fn := NewNumber(float64(2))
```
nofun97, line 13:
```suggestion
	fn := NewNumber(2)
```

PR #383, opened 2020-06-03
Rename NewMapExpr to NewDArrowExpr

orlade-anz:
This seems safe but I don't trust myself to know for sure yet.


rel/expr_darrow.go
```diff
@@ -15,7 +15,7 @@ type DArrowExpr struct {
 }
 
 // NewDArrowExpr returns a new DArrowExpr.
-func NewMapExpr(scanner parser.Scanner, lhs Expr, fn Expr) Expr {
+func NewDArrowExpr(scanner parser.Scanner, lhs Expr, fn Expr) Expr {
 	return &DArrowExpr{ExprScanner{scanner}, lhs, ExprAsFunction(fn)}
```
nofun97, line 19: seems fine, but I think `MapExpr` is a better name, what do you think @marcelocantos?

orlade-anz: Perhaps, but I would have thought at least something like SetMapExpr for consistency with the others.

marcelocantos: I chose `Map`, but I changed my mind because the term is too ambiguous.

Issue #381, opened 2020-06-03
Improve test coverage of rel/ package


orlade-anz:
As at today, the rel/ package (containing the value and expr types) has only 70% statement coverage. Since they form the foundation of arr.ai, this seems insufficient; statement coverage at least should be 100%, and we probably want more of other types as well.

orlade-anz:
So apparently the default behaviour of `go test` is to only measure coverage of a test within the same package as that test, so syntax/ won't count towards coverage of rel/. It doesn't sound like a strong ideological position, just the way it works. For some reason GoLand does the smart thing, but Coveralls just uses the basic `go test` output in the expected way.

PR #385, opened 2020-06-03
Add simple tests for accessor methods of mapping functions

orlade-anz:
- Add tests for untested methods, trivial though they may be.
- Consider just removing the methods since they're unused?

Checklist:
- [x] Added related tests


orlade-anz:
Toward #381

PR #386, opened 2020-06-03
Change optional fallback from expr to atom

marcelocantos:
Fixes # .

- Currently, `a.b?:{} => ... >> ...` parses as `a.b?:({} => ... >> ...)`, which is counterintuitive.

The expectation is that `?:` bind quite tightly, so only the `{}` is picked up as the fallback value. This is because the most common case for the `?:` fallback value is a simple "zero"-value like `0`, `[]`, `""`, etc.

Checklist:
- [ ] Added related tests


marcelocantos:
> LGTM but any reason to change it to atom?

Explained in the description.

PR #388, opened 2020-06-03
scope variables are accessible through their variable names in debug shell

nofun97:
- scope variables are accessible through their variable names in debug shell 

Checklist:
- [ ] Added related tests


docs/tutorial/shell.md
```diff
@@ -41,24 +41,27 @@ When you evaluate an arrai script and the script fails, the `arrai` program will
...
 let x = 1; (\a a + b)(x)
-@> @.x
+@> x
 1
-@> @.
-.a .x
+@> a
+1
+@> <tab>
+a x //
```
marcelocantos, line 64: \<tab> doesn't work for me. Since it's not otherwise easy to tell what names are in scope, tab-completion for globals needs to happen in or before this PR.

Issue #378, opened 2020-06-03
no error message on the wrong use of stdlib


nofun97:
Please **do not post any internal, closed source snippets** on this public issue tracker!

### Current arrai Version
There are no error message on using the stdlib properly.

```
let data = //os.file("hello.txt");
//seq.split(",", data)
```

The above script is wrong as `data` is an array of bytes and the separator must also be an array of bytes `<<",">>`. But running the script doesn't show any error message which can be misleading.

PR #393, opened 2020-06-04
#378 Checked in code to add error message.

ericzhang6222:
Fixes # .

- Show error if the delimiter in API `//seq.split(delimiter , subject)` is not an array of byte.

Checklist:
- [x] Added related tests


syntax/std_seq_bytes_helper.go
```diff
@@ -28,6 +28,9 @@ func bytesSplit(delimiter rel.Value, subject rel.Bytes) rel.Value {
 		splitted = strings.Split(subject.String(), delimiter.String())
 	case rel.GenericSet:
 		splitted = strings.Split(subject.String(), mustAsString(delimiter))
+	default:
+		panic(`the delimiter of api //seq.split(delimiter, subject) is not an array of byte, ` +
+			`it can try <<delimiter>> to generate byte array`)
```
nofun97, line 33: It needs to mention the type of the delimiter
```suggestion
		panic(fmt.Sprintf("delimiter and subject have to be of the same type, currently: delimiter: %T, subject: %T", delimiter, subject))
```

needs to import package `fmt` before merging that

PR #396, opened 2020-06-04
Support arbitrary expressions in pattern matching

### LINKED ISSUE #254: support arbitrary expressions in pattern matching
### Use-cases
The 6th step of #132.
`"hello"`, `1 + 2`: match the value of the given expression.

### LINKED ISSUE #325: Allow bare literals as value patterns
Currently, all value patterns must be enclosed in parentheses. Relax this so that literal numbers (including characters) and strings as well as `true` and `false` can be specified without thus enclosing them.

Before:
```arrai
let (42) = 42; let [x, (42)] = array; ...
x cond {(1): 42, (2): 43, [a, (0)]: a}
```

After:
```arrai
let 42 = 42; let [x, 42] = array; ...
x cond {1: 42, 2: 43, [a, 0]: a}
```

Obviously, the enclosing parentheses are still permitted, though they should be called out by a linter as unnecessary.

ChloePlanet:
Fixes #325 .
Fixes #254 .

- Support `True` and `False` in pattern matching
- Support arbitrary expressions e.g. `(1+2)`, `"hello"`

Checklist:
- [x] Added related tests

### DIFF
--- a/docs/tutorial/binding.md
+++ b/docs/tutorial/binding.md
@@ -181,13 +181,12 @@ Both let-bindings and function parameters support pattern matching. This is a
 very powerful mechanism to extract elements from a complex structure and also
 restrict what values may be used as input.
 
-The bare literals are supported:
+Bare literals are supported and will simply match their own value:
 
 ```arrai
 @> let 42 = 42; 1
 @> let "hello" = "hello"; 1
 @> let 3 = 1 + 2; 5
-@> let (1 + 2) = 3; 5
 @> let true = true; 3
 @> let true = {()}; 3
 ```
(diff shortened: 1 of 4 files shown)


docs/tutorial/binding.md
```diff
@@ -181,6 +181,17 @@ Both let-bindings and function parameters support pattern matching. This is a
 very powerful mechanism to extract elements from a complex structure and also
 restrict what values may be used as input.
 
+The bare literals are supported:
```
marcelocantos, line 184:
```suggestion
Bare literals are supported and will simply match their own value:
```

docs/tutorial/binding.md
```diff
@@ -181,6 +181,17 @@ Both let-bindings and function parameters support pattern matching. This is a
...
+
+```arrai
+@> let 42 = 42; 1
+@> let "hello" = "hello"; 1
+@> let 3 = 1 + 2; 5
+@> let (1 + 2) = 3; 5
```
marcelocantos, line 190: This isn't strictly a bare literal. I suggest an additional block with a couple of examples, with lead-in:
> Complex expressions are supported and will also match their own value. However, they must be enclosed in parentheses, `(...)`:

rel/pattern_expr.go
```diff
@@ -16,11 +16,19 @@ func NewExprPattern(expr Expr) ExprPattern {
...
+		if t == True || t == False {
+			if t.IsTrue() == value.IsTrue() {
+				return EmptyScope, nil
+			}
+			return EmptyScope, errors.Errorf("%s doesn't equal to %s", t, value)
+		}
+		return EmptyScope, fmt.Errorf("%s is not a Pattern", t)
 	default:
-		return EmptyScope, fmt.Errorf("%s is not a Pattern", p.expr)
+		return EmptyScope, fmt.Errorf("%s is not a Pattern", t)
```
marcelocantos, line 31: Can this whole function just be `p.expr.Equal(value)`?

PR #422, opened 2020-06-07
Fix expr source context

marcelocantos:
Fixes # .

- Fix source context for most expr forms and values.
- 

Checklist:
- [x] Added related tests

### DIFF
--- a/syntax/compile.go
+++ b/syntax/compile.go
@@ -35,15 +35,15 @@ var loggingOnce sync.Once
 
 // Compile compiles source string.
 func Compile(filepath, source string) (_ rel.Expr, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			if e, ok := r.(error); ok {
-				err = e
-			} else {
-				err = fmt.Errorf("error compiling %q: %v", filepath, r)
-			}
-		}
-	}()
+	// defer func() {
+	// 	if r := recover(); r != nil {
+	// 		if e, ok := r.(error); ok {
+	// 			err = e
+	// 		} else {
+	// 			err = fmt.Errorf("error compiling %q: %v", filepath, r)
+	// 		}
+	// 	}
+	// }()
 	return MustCompile(filepath, source), nil
 }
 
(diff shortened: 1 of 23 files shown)


syntax/compile.go
```diff
@@ -35,15 +35,15 @@ var loggingOnce sync.Once
...
-	}()
+	// defer func() {
+	// 	if r := recover(); r != nil {
+	// 		if e, ok := r.(error); ok {
+	// 			err = e
+	// 		} else {
+	// 			err = fmt.Errorf("error compiling %q: %v", filepath, r)
+	// 		}
+	// 	}
+	// }()
```
nofun97, line 46: removing this just makes `Compile` become `MustCompile`

syntax/compile.go
```diff
@@ -258,7 +258,8 @@ func (pc ParseContext) compileDictPattern(b ast.Branch) rel.Pattern {
 				entryPtns = append(
 					entryPtns,
 					rel.NewDictPatternEntry(
-						rel.DotIdent,
+						// TODO: ???
```
nofun97, line 261: ?

syntax/compile_scanner_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func assertCompileScanner(t *testing.T, source string) bool { //nolint:unparam
+	pc := ParseContext{SourceDir: ".."}
+	ast, err := pc.Parse(parser.NewScanner(" " + source + " "))
+	require.NoError(t, err, "%s", source)
+	expr := pc.CompileExpr(ast)
+	return assert.Equal(t, source, expr.Source().String())
+}
```
nofun97, line 17: I'm not very sure what the test does, is it supposed to see whether a literal's scanner is the whole literal?

marcelocantos: Added explanatory comments.

syntax/compile.go
marcelocantos: I've updated the message. If you know the answer, I'd like to resolve this rather than leave it as a TODO.

syntax/compile.go
marcelocantos: That was an accident. Reinstated.

PR #438, opened 2020-06-09
fix #434, adds stack frame navigation in arrai debug shell

### LINKED ISSUE #434: Stack Frame Navigation in the Debug Shell
### Use-cases
Whenever an evaluation fails, the program will drop into the debug shell with the last scope in the stack. A stack navigation solution is needed for better debugging process.

Create a command in the debug shell. `/up` and `/down` or `/u` and `/d` which navigates the stack frame and changes the global variables accordingly.

nofun97:
Fixes #434 .

- Stack frame navigation in arrai debug shell

When evaluation fails, it will drop into the debug shell with the last frame. This PR adds the `/up` which will change into the previous frame and replace the scope with the corresponding frame's scope. `/down` will change into the next frame.

Checklist:
- [x] Added related tests

### DIFF
--- a/docs/tutorial/shell.md
+++ b/docs/tutorial/shell.md
@@ -115,6 +115,10 @@ test.arrai:7:5:
 // a  b  x  y  z
 ```
 
+Stack number represents which stack is currently being used. Low stack number
+is closer to the first point of execution. The higher the stack number, the
+closer it is to the point of failure.
+
 ## Evaluating expressions
 
 To evaluate an expression in the shell, simply type it in and press enter.
--- a/docs/tutorial/shell.md
+++ b/docs/tutorial/shell.md
@@ -64,8 +64,8 @@ let x = 1; (\a a + b)(x)
 a x
 ```
 
-It is also possible to navigate through the frames by using the `/up` and
-`/down`, or `/u` and `/d` which are shorthand for the commands.
+It is also possible to navigate through the frames by using the `/up` (or
+`/u`) and `/down` (or `/d`) commands.
 
 Running the following script:
 ```arrai
--- a/docs/tutorial/shell.md
+++ b/docs/tutorial/shell.md
@@ -64,6 +64,57 @@ let x = 1; (\a a + b)(x)
 a x
 ```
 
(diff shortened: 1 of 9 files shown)


docs/tutorial/shell.md
```diff
@@ -64,6 +64,57 @@ let x = 1; (\a a + b)(x)
 a x
 ```
 
+It is also possible to navigate through the frames by using the `/up` and
+`/down`, or `/u` and `/d` which are shorthand for the commands.
```
orlade-anz, line 68:
```suggestion
It is also possible to navigate through the frames by using the `/up` (or
`/u`) and `/down` (or `/d`) commands.
```

docs/tutorial/shell.md
```diff
@@ -64,6 +64,57 @@ let x = 1; (\a a + b)(x)
...
+    b(z)(y)
+
+test.arrai:9:1:
+);
+d(x)
+
+@> <tab>
+// a  b  x  y  z
+@> /u
+2020-06-09T12:50:58.306199+10:00 INFO Stack: 0
```
orlade-anz, line 99: Very cool! Maybe explain what the stack number means. Is zero the top or bottom? Which way is up or down?

rel/expr.go
```diff
@@ -21,6 +21,10 @@ type ContextErr struct {
 	scope  Scope
 }
 
+func NewContextErr(err error, source parser.Scanner, scope Scope) ContextErr {
```
orlade-anz, line 24: ContextErr seems to have become a much more fundamental type. Could you add a brief doc explaining its purpose? It seems strange that you have to pass an array of "errors" to create a new shell instance, for example.

nofun97: added more comments on `ContextErr`. Passing an array does look strange, it's supposed to represent the frames, is it better to pass the actual error?

orlade-anz: I don't know enough about how it's used. Generally a conceptually simpler API is better, so a struct wrapping an array rather than a raw array is less to think about, but this is fine for now IMO.

PR #453, opened 2020-06-10
Implement macro support in Arrai parsing

### LINKED ISSUE #440: Support macros in Arrai

### Use-cases
Please describe the _end goal_ you are trying to achieve that has led you to request this feature.

Macros are a powerful feature in other languages (e.g. Lisp) and would allow users to easily extend the syntax of Arrai in useful, domain-specific ways.

In particular, macros provide a mechanism for:
* parsing literals into "types", e.g. `//date{2020-06-09}` -> `(year: 2020, month: 6, day: 9)`
* adding language features, e.g. `{2020 - c.dob.year for c in customer}` -> `customer => 2020 - .dob.year`.

### Attempted Solutions

Support for parsing custom grammars into ASTs gets us most of the way there, but doesn't incorporate transformation of those ASTs into meaningful Arrai data structures.

orlade-anz:
Fixes #440.

- Replace inline grammar embedding term (`{:grammar:content:}`) with upgraded macro term (`{:macro[rule]?:content:}`). `macro` can be a plain grammar AST, or a tuple containing `@grammar` and `@transform` keys. `@transform` contains functions that are applied to the `content` AST produced by parsing it with `@grammar`. As before, the output of the macro effectively replaces the `{:...:}` fragment of the program.
- If a `rule` is not explicitly provided, assume a default of the first rule in the grammar. If that fails to parse, panic.
- Similarly, if `@transform` is explicitly provided but doesn't contain a function corresponding to the name of the rule used to parse, panic.

Future work will progress the macro syntax (from `{:macro[rule]:content:}` to `macro[rule]{content}`).

Checklist:
- [x] Added related tests


orlade-anz:
A package for working with dates/times is probably a good demo of the feature, but is not included here. https://tools.ietf.org/html/rfc3339#page-12 describes a (non-official) grammar.

PR #454, opened 2020-06-10
Add shell commands to the predictions

anz-gordonj7:
Adds the shell inbuilts to the command prediction


anz-gordonj7:
@nofun97 ta, figureing out how to only show the command completions at the start of the line (if possible), updating soon

PR #483, opened 2020-06-22
Transform operators tutorial

marcelocantos:
Fixes # .

- 
- 

Checklist:
- [ ] Added related tests


docs/tutorial/transforms.md
```diff
@@ -1,3 +1,193 @@
...
+Let's explore the basic usage of the above operators, one at a time.
+
+### `->`
+
+The arrow operator, `->`, can seem almost trivial:
+
+```arrai
+@> 42 -> .
+@> 42 -> . + 1
+@> "hello, " -> . ++ "world"
```
nofun97, line 36: I think the examples need to include the resulting values. A lot of the examples here don't have results.

docs/tutorial/transforms.md
```diff
@@ -1,3 +1,193 @@
...
+```
+
+The alternative form is likewise fairly trivial:
+
+```arrai
+@> 42 -> \x x + 1
```
nofun97, line 42: Needs result. Also this can be confusing, because it seems to bind the name `x` with `42`, which wasn't explained (I thought it always binds to `.`)

docs/tutorial/transforms.md
```diff
@@ -1,3 +1,193 @@
...
+Note that `>>>` still transforms only the associated value. `@` remains
+unchanged. If you want to transform the entire tuple, use `=>` instead.
+
+### `:>`
+
+The "tuple-arrow" operator, `:>`, requires `lhs` to be a tuple. It applies the
+transform to each attribute value of `lhs` and evaluates to a tuple of the results:
+
+```arrai
+@> (r: 0.5, g: 0.2, b: 0.7) => 1 - .
```
nofun97, line 120:
```suggestion
@> (r: 0.5, g: 0.2, b: 0.7) :> 1 - .
(b: 0.3, g: 0.8, r: 0.5)
```

docs/tutorial/transforms.md
```diff
@@ -1,3 +1,193 @@
...
+
+```arrai
+@> [{1, 2}, {2, 3, 4}, {1, 5}] >> => 10 + .
+@> {(r:0.7, g:0, b:0), (r:0.4, g:0.6, b:0), (r:0.5, g:0.5, b:1)} => :> . ^ 2
+```
+
+## Interaction with order and orderby
+
+(The following discussion applies equally to `order` and `orderby`, so we'll
+only discuss `orderby`.)
```
nofun97, line 137: it might be helpful to explain the difference between `order` and `orderby`

docs/tutorial/transforms.md
```diff
@@ -1,3 +1,193 @@
...
+In order to rearrange the values accordingly, you must explicitly deal with the
+underlying representation of arrays. While this can be confusing at first, it is
+a result of consistently applying the rule that `orderby` operates on sets. It
+also gives the programmer full control over how ordering should behave. In the
+above case, a question arises: do you want two instances of the number 4 in the
+result? There isn't a single correct answer. If you want an array of all scores,
+you can do it like this:
+
+```arrai
+@> [9, 4, 2, 4] orderby [.@item, .] >> .@item
```
nofun97, line 170: maybe we should have a `sort` function for sequenced set?

docs/tutorial/transforms.md
```diff
@@ -1,3 +1,193 @@
...
+The "set-arrow" operator, `=>`, requires `lhs` to be a set. It applies the
+transform to each member of `lhs` and evaluates to a set of the results:
+
+```arrai
+@> {2, 4, 6, 8, 10} => . + 1
+@> {2, 4, 6, 8, 10} => . % 6
+```
+
+Note that, if the transform produces the same output for some of the `lhs`
+elements, the resulting set will have fewer members than `lhs`.
```
cuminandpaprika, line 67: Maybe this would be a good chance to remind a new user that sets are different from arrays, and cannot contain duplicates?

docs/tutorial/transforms.md
marcelocantos: Maybe. I'm not keen to promote extensive use of arrays. I see them as a transitional form that should eventually become of marginal utility.

PR #490, opened 2020-06-30
Protocol buffer import

### LINKED ISSUE #478: Protocol buffer import capability
Currently, we JSON-ify protobuf payloads for use in arr.ai code. JSON is somewhat lossy because it can't differentiate between struct-like and map-like data objects (think `customer` vs `map[custid]customer`), and the safe option is to map all JSON objects to dictionaries. If we support direct proto imports, we solve a few significant limitations in one go:

1. Map messages to tuples, so we can write `customer.dob` instead of `customer('dob')`.
1. Make proto ingestion safe under changes to field names.
1. Support proto map types (mapping them to dicts), including maps with non-string keys.

Rather than figuring out how to execute the `protoc` compiler, which seems to be the only supported mechanism to parse .proto files, we should assume a pb message conforming to [descriptor.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto) as the starting point for parsing a protocol buffer message.

ericzhang6222:
Fixes #478.

- Read protobuf binary message and transform to a arr.ai value, for example:
```
@> /set sysl = //encoding.proto.decode(//encoding.proto.proto, //os.file('./translate/pb/test/sysl.pb'))
@> /set decodeSyslPb = //encoding.proto.decode(sysl)
@> /set petshop = decodeSyslPb("Module" , //os.file("./translate/pb/test/petshop.pb"))
@> petshop.apps("PetShopApi").attrs('package').s
'io.sysl.demo.petshop.api'
``` 

Checklist:
- [x] Added related tests


syntax/pb_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"testing"
+
+	"github.com/arr-ai/arrai/rel"
+	"github.com/arr-ai/wbnf/parser"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTransformProtoBufToTupleFlow(t *testing.T) {
```
marcelocantos, line 11: `Protobuf`, not `ProtoBuf`

pb/pb_util.go
```diff
@@ -0,0 +1,122 @@
+package pb
+
+import (
+	"fmt"
+
+	"github.com/arr-ai/arrai/rel"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protodesc"
+	pr "google.golang.org/protobuf/reflect/protoreflect"
```
marcelocantos, line 10: Why the alias?

pb/pb_util.go
```diff
@@ -0,0 +1,122 @@
...
+	"google.golang.org/protobuf/reflect/protoregistry"
+	"google.golang.org/protobuf/types/descriptorpb"
+	"google.golang.org/protobuf/types/dynamicpb"
+)
+
+func getFileDescriptor(definition []byte) (pr.FileDescriptor, error) {
```
marcelocantos, line 16: `get` is too generic. `decode` is better.

pb/pb_util.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	rootMessageDesc := fd.Messages().ByName(pr.Name(rootMessage))
+	message := dynamicpb.NewMessage(rootMessageDesc)
+	err = proto.Unmarshal(data, message)
+	if err != nil {
+		return nil, err
+	}
+
+	tuple, err := walkThroughMessage(pr.ValueOf(message.ProtoReflect()))
```
marcelocantos, line 48: `walkThroughMessage` only describes the traversal mechanism. A function's name should primarily describe what it achieves. How it does it is a secondary consideration and should only be part of the name if the caller cares about this aspect.

Call it `valueFromMessage`.

pb/pb_util.go
```diff
@@ -0,0 +1,122 @@
+package pb
```
marcelocantos, line 1: Move the whole package to `translate/pb`.

pb/pb_util.go
ericzhang6222: It is shorter and simpler than `protoreflect `

pb/pb_util.go
ericzhang6222: I think `walkThroughMessageToBuildValue` is better.

pb/pb_util.go
marcelocantos: It doesn't occur often enough for the indirection to be warranted. I had to lookup what `pr` meant, where `protoreflect` would be obvious.

syntax/pb_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	assert.NoError(t, err)
+	tuple, _ := val.(rel.Tuple)
+	apps, _ := tuple.Get("apps")
+
+	assert.NotNil(t, apps)
+}
+
+func TestTransformProtoBufToTupleCompareResult(t *testing.T) {
+	code := "//encoding.proto.decode(//os.file('../pb/test/sysl.pb'), //os.file('../pb/test/petshop'), 'Module')"
```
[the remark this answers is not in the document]
ericzhang6222: Do you want decoding them to `rel.Value`? The decoding of protobuf bytes is `FileDescriptor`, and if want to convert it to an arr.ai type `rel.Value`. It has to add a wrapper which implements interface `rel.Value` and it looks like:
```
type FileDescriptorWrapper struct {
	descriptor pr.FileDescriptor
}

func (descriptor FileDescriptorWrapper) Kind() int {
	return 1
}

func (descriptor FileDescriptorWrapper) Equal(desc interface{}) bool {
	return false
}

func (descriptor FileDescriptorWrapper) Eval(scope rel.Scope) (rel.Value, error) {
	return nil, nil
}

func (descriptor FileDescriptorWrapper) Export() interface{} {
	return nil
}

func (descriptor FileDescriptorWrapper) Hash(seed uintptr) uintptr {
	return 0
}

func (descriptor FileDescriptorWrapper) IsTrue() bool {
	return false
}

func (descriptor FileDescriptorWrapper) Less(v rel.Value) bool {
	return false
}

func (descriptor FileDescriptorWrapper) Negate() rel.Value {
	return nil
}

func (descriptor FileDescriptorWrapper) Source() parser.Scanner {
	return parser.Scanner{}
}

func (descriptor FileDescriptorWrapper) String() string {
	return ""
}
```
Most of them will be empty methods, and it looks odd. Also it has to change some other methods, for example:
```
func reprValue(v Value, w io.Writer) {
    switch v := v.(type) {
    case String:
        reprString(v, w)
    case Bytes:
        reprBytes(v, w)
    case Array:
        reprArray(v, w)
    case Dict:
        reprDict(v, w)
    case GenericSet:
        reprSet(v, w)
    case StringCharTuple:
        reprStringCharTuple(v, w)
    case ArrayItemTuple:
        reprArrayItemTuple(v, w)
    case DictEntryTuple:
        reprDictEntryTuple(v, w)
    case *GenericTuple:
        reprTuple(v, w)
    case Number:
        reprNumber(v, w)
    default:
        panic(fmt.Errorf("Repr(): unexpected Value type %T: %[1]v", v)) //nolint:golint
    }
}
```
I think this will add unnecessary complexity.

marcelocantos:
RE: https://github.com/arr-ai/arrai/pull/490#discussion_r449831079

No, you don't have to wrap a `FileDescriptor`. See https://github.com/arr-ai/arrai/issues/478#issuecomment-650970308 for details. In summary, the first parameter to `//encoding.proto.decode` is a tuple representing a `FileDescriptor`. It is in fact the output from a call to `//encoding.proto.decode(//encoding.proto.proto, somePbBytes)`, which means you already have code that produces the `rel.Value` you need as input.

ericzhang6222:
> RE: [#490 (comment)](https://github.com/arr-ai/arrai/pull/490#discussion_r449831079)
> 
> No, you don't have to wrap a `FileDescriptor`. See [#478 (comment)](https://github.com/arr-ai/arrai/issues/478#issuecomment-650970308) for details. In summary, the first parameter to `//encoding.proto.decode` is a tuple representing a `FileDescriptor`. It is in fact the output from a call to `//encoding.proto.decode(//encoding.proto.proto, somePbBytes)`, which means you already have code that produces the `rel.Value` you need as input.

Actually, I got this point. The issue is how to use a tuple representing a `FileDescriptor`. The method which creates a tuple like `rel.NewTuple(attrs)` and `rel.NewTupleFromMap(m)` have to involve `rel.Value` always. I checked possible methods, they has to add a `rel.Value` to tuple or tuple attr. That's why I think it must to add a wrapper for `FileDescriptor`, and looks like it is not necessary. Can you advice more details if I was wrong?

translate/pb/pb_util.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protodesc"
+	pr "google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// DecodeFileDescriptor parses protobuf definition file and decodes to file descriptor
+func DecodeFileDescriptor(definition []byte) (pr.FileDescriptor, error) {
```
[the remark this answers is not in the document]
ericzhang6222: As breaking logic to 3 different functions in code `var stdProtobufDecode = rel.NewNativeFunction` of file `std_encoding.go`. And this code is in package syntax. 
It can change back and then will export `valueAsBytes` and `valueAsString` in package `syntax` as they are used in `var stdProtobufDecode = rel.NewNativeFunction`.

syntax/std_encoding.go
```diff
@@ -2,8 +2,28 @@ package syntax
...
 func stdEncoding() rel.Attr {
-	return rel.NewTupleAttr("encoding", stdEncodingJSON())
+	return rel.NewTupleAttr("encoding",
+		stdEncodingJSON(),
+		stdEncodingProtoProto(),
+		stdEncodingProto(),
+	)
+}
+
+// this is a placeholder to represent //encoding.proto.proto
```
orlade-anz, line 16: Nit: explain here why it's a placeholder (i.e. it simply indicates to the decode function to decode a message to the standard proto format, or something)

tools/value_util.go
```diff
@@ -0,0 +1,31 @@
+package tools
+
+import "github.com/arr-ai/arrai/rel"
+
+func ValueAsString(v rel.Value) (string, bool) {
```
orlade-anz, line 5: Add comments to describe these functions. It's relatively clear what they do, but it's not clear why - why would I call `ValueAsString` and not just `value.String()`?

translate/pb/pb_util.go
```diff
@@ -0,0 +1,163 @@
...
+	// Initialize the File descriptor object
+	fd1, err := protodesc.NewFile(file1, protoregistry.GlobalFiles)
+	if err != nil {
+		return nil, err
+	}
+
+	return fd1, nil
+}
+
+// WalkThroughMessageToBuildValue walks through protobuf message and build a value whose type is rel.Value
```
orlade-anz, line 36: Nit: Naming: while you do use this function to walk through the message, it's actually more general than that. Looking at the signature, it's converting a `protoreflect.Value` into a `rel.Value`. In the case where the input is a Message, it will walk through it, but in other cases it doesn't.

Maybe something like `ConvertProtoValueToSysl`?

translate/pb/pb_util.go
```diff
@@ -0,0 +1,163 @@
...
+func WalkThroughMessageToBuildValue(val protoreflect.Value) (rel.Value, error) {
+	switch message := val.Interface().(type) {
+	case protoreflect.Message:
+		attrs := []rel.Attr{}
+		message.Range(func(desc protoreflect.FieldDescriptor, val protoreflect.Value) bool {
+			item, err := WalkThroughMessageToBuildValue(val)
+			if err != nil {
+				attrs = append(attrs, rel.NewAttr("@error_"+string(desc.Name()),
+					rel.NewString([]rune(err.Error()))))
```
orlade-anz, line 45: Is this necessary? If we fail to convert the value to a `rel.Value`, it seems as though the whole function should error out. What's the case where a conversion error would be acceptable?

This would simplify all arms of this switch as well.

translate/pb/pb_util.go
```diff
@@ -0,0 +1,163 @@
...
+									rel.NewString([]rune(string(name)))))
+						} else {
+							attrs = append(attrs, rel.NewAttr(string(desc.Name()), item))
+						}
+					default:
+						attrs = append(attrs, rel.NewAttr(string(desc.Name()), item))
+					}
+				} else {
+					attrs = append(attrs, rel.NewAttr(string(desc.Name()), item))
+				}
```
orlade-anz, line 64: Style nit: It's difficult to keep all of the context of such a deep nested conditional in your head at once. Rule of thumb, handle the simplest/most common cases first, then follow up with the edge cases. Consider something more like:

```go
if !enum || item.(type) != number
  append string name
  continue

// handle numbers
```

Perhaps even pull out a function like `getName(desc, item) string`? Might be overkill, but getting 9 tabs in is probably too many.

translate/pb/pb_util.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		return val, nil
+	default:
+		// []byte and protoreflect.Enum etc.
+		return rel.NewTuple(rel.NewStringAttr("@error",
+			[]rune(fmt.Errorf("%T is not supported data type", message).Error()))), nil
+	}
+}
+
+//StdProtobufDecode transforms the protocol buffer message to a tuple.
```
orlade-anz, line 125: Nit: Space after //
```suggestion
// StdProtobufDecode transforms the protocol buffer message to a tuple.
```

syntax/pb_test.go
```diff
@@ -0,0 +1,65 @@
...
+	val, err := codeExpr.Eval(rel.EmptyScope)
+
+	assert.NoError(t, err)
+	shop, _ := val.(rel.Tuple)
+	apps, _ := shop.Get("apps")
+
+	assert.NotNil(t, apps)
+}
+
+func TestTransformProtobufToTupleCompareResult(t *testing.T) {
```
orlade-anz, line 30: Nit: put each of these cases in a separate function if they are indeed testing different things, and include a little bit in the function name to describe the case being tested.

translate/pb/pb_util.go
```diff
@@ -0,0 +1,163 @@
+package pb
```
orlade-anz, line 1: General nit: It's important for reading and understanding the code that the filename describe the code it contains. "pb_util" just says "something to do with protobufs". This is both unhelpful for a reader of the code, and encourages people to dump anything related to protobufs into this file.

Consider a name that describes code that does exactly one thing, e.g. "pb_decoder".

tools/value_util.go
ericzhang6222: As it has to deal with type `rel.GenericSet`. In arr.ai, all empty sets are the same and `""`, `{}` and `[]` will be parsed to `rel.GenericSet`.

translate/pb/pb_util.go
ericzhang6222: Changed to pb_decoder.go

translate/pb/pb_util.go
ericzhang6222: It can provide all errors in the `rel.Value` in one go.

translate/pb/pb_decoder_entry.go
```diff
@@ -0,0 +1,76 @@
...
+
+// StdProtobufDecoder transforms the protocol buffer message to a tuple.
+var StdProtobufDecoder = rel.NewNativeFunction("decode", func(tupleParam rel.Value) (rel.Value, error) {
+	tuple, isTuple := tupleParam.(rel.Tuple)
+	if !isTuple {
+		return nil, fmt.Errorf("//encoding.proto.decode: tupleParam not tuple")
+	}
+
+	if !tuple.IsTrue() {
+		// the call looks like `/set sysl = //encoding.proto.decode(//encoding.proto.proto, //os.file("sysl.pb"))`
```
orlade-anz, line 21: Nit: As below

translate/pb/pb_decoder_entry.go
```diff
@@ -0,0 +1,76 @@
...
+			if !is {
+				return nil, fmt.Errorf("//encoding.proto.decode: definition not bytes")
+			}
+
+			return rel.NewTuple(rel.NewAttr("fileDescriptor", definition)), nil
+		}), nil
+	}
+
+	// the call looks like `/set decodeSyslPb = //encoding.proto.decode(sysl)`
+	// after `/set sysl = //encoding.proto.decode(//encoding.proto.proto, //os.file("sysl.pb"))`
```
orlade-anz, line 33: Nit: This kind of comment would be useful on the function since it describes usage; it's less useful inside since it doesn't describe the implementation.

Also I think as an example it would be better as a block of plain arr.ai without the `/set`, something like:

```
let sysl = //encoding.proto.decode(//encoding.proto.proto, //os.file("sysl.pb"));
let petshop = //encoding.proto.decode(sysl, 'Module', //os.file("petshop.pb"));
petshop.apps('PetShopApi')
```

syntax/pb_test.go
```diff
@@ -0,0 +1,61 @@
...
+	AssertCodesEvalToSameValue(t, `'PetShopApi'`, code)
+
+	code = sharedCode + `shop.apps('PetShopApi').attrs('package').s`
+	AssertCodesEvalToSameValue(t, "'io.sysl.demo.petshop.api'", code)
+
+	code = sharedCode + `shop.apps('PetShopApi').endpoints('GET /petshop').attrs('patterns').a.elt(0).@item.s`
+	AssertCodesEvalToSameValue(t, "'rest'", code)
+
+	code = sharedCode + `shop.apps('PetShopApi').endpoints('GET /petshop').attrs('patterns').a.elt.@item.@item.s`
+	AssertCodesEvalToSameValue(t, "'rest'", code)
```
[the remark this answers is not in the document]
ericzhang6222: A flat array can't get element index by `@`, for example: `petshop.apps('PetShopApi').endpoints('GET /petshop').attrs('patterns').a.elt.@` will print indedx `0`.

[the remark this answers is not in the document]
orlade-anz: Sure, but `a.elt.@item.s` should work, right? Since there's only one element.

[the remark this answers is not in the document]
ericzhang6222: Now code transforms list to a tuple array, so it is a tuple array still even it has only one element. That's why it can use `a.elt.@item.@item.s`, and actually it can run `a.elt(0).@item.s` to access too. So I think `a.elt(0).@item.s` can work for you too.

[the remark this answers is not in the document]
orlade-anz: Right, but what I'm saying is there should only be one level of array in this case.

- `a` is an `Array` message, so it's a tuple with a `part` field.
- `part` is a `repeated string` field so it should be an array, not an array of arrays.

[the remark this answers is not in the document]
orlade-anz: Sorry, yes, I just read the proto wrong: `a` is an `Array` with repeated `Attribute`s, each of which is a `repeated string`, so it is indeed an array of arrays, not a flat array. Carry on!

syntax/pb_test.go
```diff
@@ -0,0 +1,67 @@
+package syntax
```
orlade-anz, line 1: Nit: It feels like this test should be in `translate/pb/test`, alongside all the files it imports?

translate/pb/pb_decoder_entry.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if !tuple.IsTrue() {
+		// the call looks like `let sysl = //encoding.proto.decode(//encoding.proto.proto, //os.file('sysl.pb'))`
+		return rel.NewNativeFunction("decode$1", func(definition rel.Value) (rel.Value, error) {
+			_, is := tools.ValueAsBytes(definition)
+			if !is {
+				return nil, fmt.Errorf("//encoding.proto.decode: definition not bytes")
+			}
+
+			return rel.NewTuple(rel.NewAttr("fileDescriptor", definition)), nil
```
orlade-anz, line 33: Why do we need this `(fileDescriptor: *)` boilerplate, rather than just returning `definition`? It introduces an extra level of abstraction with the concept of "a structure that contains a fileDescriptor", but I don't see the benefit of doing so.

syntax/pb_test.go
ericzhang6222: As `AssertCodesEvalToSameValue` now is in package `syntax`, so moving this file to package `pb` will causing cycle import which is not allowed in go. I will raise a new PR to update after this one.

translate/pb/pb_decoder_entry.go
ericzhang6222: `rel.NewTuple(rel.NewAttr("fileDescriptor", definition))` is a tuple which has `sysl.proto` definition, which can support the syntax:
```
@> /set sysl = //encoding.proto.decode(//encoding.proto.proto, //os.file("sysl.pb"))
@> /set decodeSyslPb = //encoding.proto.decode(sysl)
```
the first argument of `//encoding.proto.decode` is an arr.ai tuple. If it return definition directly, it has to check data type which is `rel.Tuple` or `rel.Bytes` to see now it is the first call with `//encoding.proto.proto` as the first argument or second call.
Now it make sure the first argument is `rel.Tuple` and there is a check:
```
if !isTuple {
		return nil, fmt.Errorf("//encoding.proto.decode: tupleParam not tuple")
	}
```

PR #502, opened 2020-07-14
Implements merge operator `+>`

### LINKED ISSUE #133: Merge operator

### Use-cases

Merging tuples and functions is desirable. Merging two tuples combines all rhs attributes and any lhs attributes whose name isn't on the rhs into a new tuple. For example, merging `(a: 1, b: 2)` and `(b: 3, c: 4)` produces `(a: 1, b: 3, c: 4)`.

This is already implemented as `+`, but this operator should be strictly reserved for its common numeric interpretation. In future, libraries should be able to implement arithmetic over tuples, e.g.: `(x: 1, y: 2) + (x: 3, y: -1) = (x: 4, y: 1)`, which would conflict with the current `+` semantic.

### Proposed Solution

The proposed syntax is `a +> b`. The syntactic asymmetry emphasises the semantic asymmetry (b's attributes take priority over a's).

### Alternatives Considered

`a +> b` is slightly confusing since the directionality might give the impression that a's attributes override b's.

| syntax | notes |
|:-|:-|
| `a <+ b` | Infeasible, since it currently parses as `a < +b`. |
| `a +< b` | Confusing. Doesn't look like a single operator. |
| `a + b` | As discussed above, this should be reserved for the common arithmetic meaning. |
| `a ++ b` | Applies to special sets (arrays, strings, ...) and has a different semantic (applies an offset to every `@` on the RHS). |
| `a merge b` | Adds a keyword and is a bit verbose for what is likely to be a very common operation, especially considering #485. |

### Possible extension to dictionaries

Dictionaries have similar behaviour to tuples. `+>` could also be made to work for them. This technically violates the principle of syntactic transparency (what you get can be inferred directly from what you see). But tuples and dicts are close enough cousins that it's probably OK to make an exception for them.

ericzhang6222:
Fixes #133 .

- Merges two tuples and combines all rhs(right-hand side) attributes and any lhs(left-hand side) attributes whose name isn't on the rhs into a new tuple.
- 

Checklist:
- [x] Added related tests

### DIFF
--- a/docs/example.md
+++ b/docs/example.md
@@ -262,3 +262,19 @@ $ arrai eval '{|a,b| (1,2), (1,3)} <&> {|a,c| (1,2), (1,3)}'
 $ arrai eval '{|a,b,c| (1,2,2), (1,2,3)} <&> {|b,c,d| (2,3,4), (1,3,4)}'
 {(a: 1, b: 2, c: 3, d: 4)}
 ```
+
+### Merging operator
+
+Merges two tuples and combines all rhs(right-hand side) attributes and any lhs(left-hand side) attributes whose name isn't on the rhs into a new tuple.
+
+#### Examples
+
+```bash
+arrai e '(a: 1, b: 2) +> (b: 3, c: 4)'
+(a: 1, b: 3, c: 4)
+```
+
+```bash
+arrai e '(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)'
+(a: 1, b: (c: 4), c: 4)
+```
--- a/docs/example.md
+++ b/docs/example.md
@@ -263,18 +263,23 @@ $ arrai eval '{|a,b,c| (1,2,2), (1,2,3)} <&> {|b,c,d| (2,3,4), (1,3,4)}'
 {(a: 1, b: 2, c: 3, d: 4)}
 ```
 
-### Merging operator
+### Merge
 
-Merges two tuples and combines all rhs(right-hand side) attributes and any lhs(left-hand side) attributes whose name isn't on the rhs into a new tuple.
+Merge combines two tuples, producing a single tuple containing a union of their attributes. If the same name is present in both the LHS (left-hand side) and RHS (right-hand side) tuples, the RHS value takes precedence in the output.
 
 #### Examples
 
 ```bash
-arrai e '(a: 1, b: 2) +> (b: 3, c: 4)'
+$ arrai e '(a: 1, b: 2) +> (b: 3, c: 4)'
 (a: 1, b: 3, c: 4)
 ```
 
 ```bash
-arrai e '(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)'
+$ arrai e '(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)'
 (a: 1, b: (c: 4), c: 4)
+```
+
+```bash
+$ arrai e '(a: (b: 1)) +> (a: (c: 2))'
+(a: (c: 2))
 ```
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -298,7 +298,7 @@ func (e *BinExpr) Eval(local Scope) (_ Value, err error) {
 	return val, nil
 }
 
-// evalValForAddArrow evaluates for operator `+>`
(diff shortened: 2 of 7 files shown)


docs/example.md
```diff
@@ -262,3 +262,19 @@ $ arrai eval '{|a,b| (1,2), (1,3)} <&> {|a,c| (1,2), (1,3)}'
...
 ```
+
+### Merging operator
+
+Merges two tuples and combines all rhs(right-hand side) attributes and any lhs(left-hand side) attributes whose name isn't on the rhs into a new tuple.
+
+#### Examples
+
+```bash
+arrai e '(a: 1, b: 2) +> (b: 3, c: 4)'
```
orlade-anz, line 273:
```suggestion
$ arrai e '(a: 1, b: 2) +> (b: 3, c: 4)'
```

docs/example.md
```diff
@@ -262,3 +262,19 @@ $ arrai eval '{|a,b| (1,2), (1,3)} <&> {|a,c| (1,2), (1,3)}'
...
+(a: 1, b: 3, c: 4)
+```
+
+```bash
+arrai e '(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)'
```
orlade-anz, line 278:
```suggestion
$ arrai e '(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)'
```

docs/example.md
```diff
@@ -262,3 +262,19 @@ $ arrai eval '{|a,b| (1,2), (1,3)} <&> {|a,c| (1,2), (1,3)}'
...
+(a: 1, b: (c: 4), c: 4)
```
orlade-anz, line 279: This is a nice example, but it leaves it a little ambiguous as to whether the merge is "deep" (i.e. recursive) or "shallow" (i.e. just replace the whole attribute value).

For example:

```
$ arrai e '(a: (b: 1)) +> (a: (c: 2))'
```

Would this produce `(a: (b: 1, c: 2))` or `(a: (c: 2))`? I'm guessing the latter but ideally the docs would tell me so I don't have to guess.

docs/example.md
```diff
@@ -262,3 +262,19 @@ $ arrai eval '{|a,b| (1,2), (1,3)} <&> {|a,c| (1,2), (1,3)}'
 $ arrai eval '{|a,b,c| (1,2,2), (1,2,3)} <&> {|b,c,d| (2,3,4), (1,3,4)}'
 {(a: 1, b: 2, c: 3, d: 4)}
 ```
+
+### Merging operator
```
orlade-anz, line 266: Nit: For consistency with above operators.

```suggestion
### Merge
```

docs/example.md
```diff
@@ -262,3 +262,19 @@ $ arrai eval '{|a,b| (1,2), (1,3)} <&> {|a,c| (1,2), (1,3)}'
...
+
+Merges two tuples and combines all rhs(right-hand side) attributes and any lhs(left-hand side) attributes whose name isn't on the rhs into a new tuple.
```
orlade-anz, line 268: Small nit: good description but a bit of a mouthful. Two sentences is clearer, saying a) what it does, and b) addressing edge cases.

```suggestion
Merge combines two tuples, producing a single tuple containing a union of their attributes. If the same name is present in both the LHS (left-hand side) and RHS (right-hand side) tuples, the RHS value takes precedence in the output.
```

rel/expr_binary.go
```diff
@@ -80,6 +80,14 @@ func NewAddExpr(scanner parser.Scanner, a, b Expr) Expr {
 		})
 }
 
+// NewAddArrowExpr returns a new BinExpr which supports operator `+>`
```
orlade-anz, line 83: Nit: full stop.
```suggestion
// NewAddArrowExpr returns a new BinExpr which supports operator `+>`.
```

rel/expr_binary.go
```diff
@@ -289,3 +297,16 @@ func (e *BinExpr) Eval(local Scope) (_ Value, err error) {
 	}
 	return val, nil
 }
+
+// evalValForAddArrow evaluates operator `+>`
```
orlade-anz, line 301: Nit: full stop.
```suggestion
// evalValForAddArrow evaluates operator `+>`.
```

PR #505, opened 2020-07-15
Merge operator

### LINKED ISSUE #133: Merge operator

ericzhang6222:
Fixes #133  .

- Support Dict
- 

Checklist:
- [x] Added related tests

### DIFF
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -298,7 +298,7 @@ func (e *BinExpr) Eval(local Scope) (_ Value, err error) {
 	return val, nil
 }
 
-// evalValForAddArrow evaluates for operator `+>`
+// evalValForAddArrow evaluates operator `+>`
 func evalValForAddArrow(lhs, rhs Value) (Value, error) {
 	if lhs, ok := lhs.(Tuple); ok {
 		if rhs, ok := rhs.(Tuple); ok {
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 
+	"github.com/arr-ai/frozen"
 	"github.com/arr-ai/wbnf/parser"
 	"github.com/go-errors/errors"
 )
@@ -306,7 +307,28 @@ func evalValForAddArrow(lhs, rhs Value) (Value, error) {
 		}
 	}
 
+	if lhs, ok := lhs.(Dict); ok {
+		if rhs, ok := rhs.(Dict); ok {
+			return mergeDicts(lhs, rhs), nil
+		}
+	}
+
 	return nil, errors.Errorf(
-		"Both args to %q must be tuples, not %T and %T",
+		"Both args to %q must be tuples or dicts, not %T and %T",
 		"+>", lhs, rhs)
+}
+
+func mergeDicts(lhs Dict, rhs Dict) Dict {
+	tempMap := lhs.m
+
+	for e := rhs.DictEnumerator(); e.MoveNext(); {
+		key, value := e.Current()
+
+		if lhs.m.Has(key) {
+			tempMap = tempMap.Without(frozen.NewSet(key))
+		}
+		tempMap = tempMap.With(key, value)
+	}
+
+	return Dict{m: tempMap}
 }
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -80,7 +80,7 @@ func NewAddExpr(scanner parser.Scanner, a, b Expr) Expr {
 		})
 }
 
-// NewAddArrowExpr returns a new ArrowExpr.
+// NewAddArrowExpr returns a new BinExpr which supports operator `+>`
 func NewAddArrowExpr(scanner parser.Scanner, lhs, rhs Expr) Expr {
 	return newBinExpr(scanner, lhs, rhs, "+>", "(%s +> %s)",
 		func(lhs, rhs Value, _ Scope) (Value, error) {
--- a/rel/expr_binary.go
+++ b/rel/expr_binary.go
@@ -307,6 +307,6 @@ func evalValForAddArrow(lhs, rhs Value) (Value, error) {
 	}
 
 	return nil, errors.Errorf(
-		"Both args to %q must be tuples or dicts, not %T and %T",
+		"Both args to %q must be tuples, not %T and %T",
 		"+>", lhs, rhs)
 }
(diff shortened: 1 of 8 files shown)


rel/expr_binary.go
```diff
@@ -306,7 +307,28 @@ func evalValForAddArrow(lhs, rhs Value) (Value, error) {
...
 
+	if lhs, ok := lhs.(Dict); ok {
+		if rhs, ok := rhs.(Dict); ok {
+			return mergeDicts(lhs, rhs), nil
+		}
+	}
+
 	return nil, errors.Errorf(
-		"Both args to %q must be tuples, not %T and %T",
+		"Both args to %q must be tuples or dicts, not %T and %T",
```
orlade-anz, line 317: Nit: presumably you can't merge a dict into a tuple
```suggestion
		"Both args to %q must be both tuples or both dicts, not %T and %T",
```

syntax/expr_binary_test.go
```diff
@@ -19,13 +19,19 @@ func TestRelationCall(t *testing.T) {
 
 func TestOpsAddArrow(t *testing.T) {
 	t.Parallel()
+	// Tuples
```
orlade-anz, line 22: Please create separate test `func`s for different test cases instead of mixing them in one.

syntax/expr_binary_test.go
```diff
@@ -19,13 +19,19 @@ func TestRelationCall(t *testing.T) {
...
-	AssertCodesEvalToSameValue(t, `(a: 1, b: 2) +> (b: 3, c: 4)`, `(a: 1, b: 3, c: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 2)) +> (b: 3, c: 4)`, `(a: 1, b: 3, c: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)`, `(a: 1, b: (c: 4), c: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 2)) +> (a: (b: 1), b: (c: 4))`, `(a: (b: 1), b: (c: 4))`)
+	// Dicts
+	AssertCodesEvalToSameValue(t, `{'a': 1, 'b': 3, 'd': 4}`, `{'a': 1, 'b': 2} +> {'b': 3, 'd': 4}`)
+	AssertCodesEvalToSameValue(t, `{'a': 1, 'b': 3, 'd': 4}`, `{'a': 1, 'b': {'a': 2}} +> {'b': 3, 'd': 4}`)
+	AssertCodesEvalToSameValue(t, `{'a': {'c': 2}}`, `{'a': {'b': 1}} +> {'a': {'c': 2}}`)
 
+	//
```
orlade-anz, line 34: These can be a separate case as well

orlade-anz: Also, maybe add a test for lists and strings? They're more dict-like than sets and ints

rel/expr_binary.go
```diff
@@ -306,7 +307,28 @@ func evalValForAddArrow(lhs, rhs Value) (Value, error) {
...
+func mergeDicts(lhs Dict, rhs Dict) Dict {
+	tempMap := lhs.m
+
+	for e := rhs.DictEnumerator(); e.MoveNext(); {
+		key, value := e.Current()
+
+		if lhs.m.Has(key) {
+			tempMap = tempMap.Without(frozen.NewSet(key))
+		}
+		tempMap = tempMap.With(key, value)
```
nofun97, line 330: I don't think you need to do a `Without` and then a `With` operation. Just `With` should do it, does it not work?

syntax/expr_binary_test.go
```diff
@@ -17,15 +17,27 @@ func TestRelationCall(t *testing.T) {
...
+	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 4), c: 4)`, `(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)`)
+	AssertCodesEvalToSameValue(t, `(a: (b: 1), b: (c: 4))`, `(a: 1, b: (c: 2)) +> (a: (b: 1), b: (c: 4))`)
+}
 
-	AssertCodesEvalToSameValue(t, `(a: 1, b: 2) +> (c: 3, d: 4)`, `(a: 1, b: 2, c: 3, d: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: 2) +> (b: 3, c: 4)`, `(a: 1, b: 3, c: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 2)) +> (b: 3, c: 4)`, `(a: 1, b: 3, c: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 2)) +> (b: (c: 4), c: 4)`, `(a: 1, b: (c: 4), c: 4)`)
-	AssertCodesEvalToSameValue(t, `(a: 1, b: (c: 2)) +> (a: (b: 1), b: (c: 4))`, `(a: (b: 1), b: (c: 4))`)
+func TestOpsAddArrowForDicts(t *testing.T) {
```
nofun97, line 29: do another test where both `lhs` and `rhs` have no mutual keys. Do the same for tuples.

PR #508, opened 2020-07-16
Add std YAML decoding and yaml/yx transform command

### LINKED ISSUE #507: YAML transform CLI command

### Purpose
Please describe the _end goal_ you are trying to achieve that has led you to request this feature.

I want to convert YAML to arr.ai.

### Suggested approaches
What have you tried, and how might this problem be solved?

- `arrai jx` already exists; the same should exist for YAML.

orlade-anz:
Fixes #507.

- Add `//encoding.yaml.decode` to std library. Virtually identical to `//encoding.json.decode`.
- Add `yaml`/`yx` command to transform YAML. Virtually identical to `json`/`jx` command.

Checklist:
- [x] Added related tests


cmd/arrai/yaml.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/arr-ai/arrai/rel"
+	"github.com/arr-ai/arrai/translate"
+	"github.com/urfave/cli/v2"
+)
+
+var yamlCommand = &cli.Command{
```
anzdaddy, line 13: I don’t know how useful it is to have a command to translate yaml to arr.ai syntax. In practice, I think the intent will almost always be to do something with the translated values rather than just output them to stdout. On that note, `//encoding.yaml.decode` is certainly useful. In fact, `arrai yx` is easily expressible as `arrai eval '//encoding.yaml.decode(//os.stdin)'`.

To address the obvious question, I have similar reservations about the `jx` command. We don't it either.

orlade-anz: Agree. Removed yx, kept decode.

PR #511, opened 2020-07-17
Pretty printer

### LINKED ISSUE #171: Pretty printer
### Tasks
1. [ ] Add a pretty print function to arrai itself. `//fmt.pretty(config, value)`

2. [ ] Using `$"..."` instead of regular string to show indentations and newlines.
For example, currently the output `//net.http.get` uses regular strings, which can return strings such as `"random\nstring"`. The use of `$"..."` can show the indentations.
```
(
  body: $"
    all: test lint wasm
    # TODO: If this Makefile is ever used for CI, suppress timingsensitive there.
    test:
    	go test $(GOTESTFLAGS) -tags timingsensitive ./...
    lint:
    	golangci-lint run
    wasm:
    	GOOS=js GOARCH=wasm go build -o /tmp/arrai.wasm ./cmd/arrai",
  header:(
...
```

3. [ ] Pretty print tuple
```
  header:(
    'Accept-Ranges':               ['bytes'],
    'Access-Control-Allow-Origin': ['*'],
    'Cache-Control':               ['max-age=300'],
    'Connection':                  ['keep-alive'],
    'Content-Security-Policy':     ["default-src 'none'; style-src 'unsafe-inline'; sandbox"],
    'Content-Type':                ['text/plain; charset=utf-8'],
    'Date':                        ['Mon, 13 Apr 2020 12:17:00 GMT'],
    'Etag':                        ['W/"d7460fe5d998b2f25ba976ca6ef6646215c0b3314608fb5462a48f6412334f13"'],
    'Expires':                     ['Mon, 13 Apr 2020 12:22:00 GMT'],
    'Source-Age':                  ['246'],
    'Strict-Transport-Security':   ['max-age=31536000'],
    'Vary':                        ['Authorization,Accept-Encoding'],
    'Via':                         ['1.1 varnish (Varnish/6.0)', '1.1 varnish'],
    'X-Cache':                     ['MISS, HIT'],
    'X-Cache-Hits':                ['0, 1'],
    'X-Content-Type-Options':      ['nosniff'],
    'X-Fastly-Request-Id':         ['fe13435de8287048b85b80e6057b941d3470dea6'],
    'X-Frame-Options':             ['deny'],
    'X-Github-Request-Id':         ['8956:2496:13C90C:1631AA:5E945745'],
    'X-Served-By':                 ['cache-mel19034-MEL'],
    'X-Timer':                     ['S1586780220.448718,VS0,VE1'],
    'X-Xss-Protection':            ['1; mode=block']
  )
```

ericzhang6222:
Fixes #171  .

- Add a pretty print function to arrai itself. `//fmt.pretty(value)` which can print more readable format of string represents `rel.Value`:
```
{
  b: 2,
  a: 1,
  c: (
    d: 11,
    f: {
      111,
      222
    },
    e: 22
  )
}
```

Checklist:
- [x] Added related tests


syntax/std_fmt_test.go
```diff
@@ -0,0 +1,24 @@
...
+func TestFmtPrettyTuple(t *testing.T) {
+	t.Parallel()
+	AssertCodesEvalToSameValue(t, `""`, `//fmt.pretty((a:1,b:2,c:3))`)
+	AssertCodesEvalToSameValue(t, `""`, `//fmt.pretty((a:1,b:2,c:(d:11,e:22)))`)
+}
+
+func TestFmtPrettyMixed(t *testing.T) {
+	t.Parallel()
+	AssertCodesEvalToSameValue(t, `""`, `//fmt.pretty({'a':1,'b':2,'c':(d:11, e:22)})`)
+	AssertCodesEvalToSameValue(t, `""`, `//fmt.pretty((a:1, b:2, c:{'d':11, 'e':22}))`)
```
orlade-anz, line 23: This doesn't look right; `pretty` shouldn't eval to `""`?

syntax/format.go
```diff
@@ -0,0 +1,115 @@
...
+		format = format + "\n"
+	}
+
+	return format
+}
+
+func getIdents(identsNum int) string {
+	var sb strings.Builder
+	for i := 0; i < identsNum; i++ {
+		sb.WriteString("\t")
```
orlade-anz, line 111: Nit: Can we use two spaces rather than a tab? arr.ai data can get deeply nested quickly, and once the indents get longer than a screen and have to wrap the pretty printing will be almost useless, so two spaces should keep the indents smaller.

syntax/format.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/arr-ai/arrai/rel"
+)
+
+// FormatString returns a string which represents `rel.Value` with more reabable format.
+// For example, `{b: 2, a: 1, c: (a: 2, b: {aa: {bb: (a: 22, d: {3, 1, 2})}})}` can be formatted to:
```
orlade-anz, line 11: Nit
```suggestion
// For example, `{b: 2, a: 1, c: (a: 2, b: {aa: {bb: (a: 22, d: {3, 1, 2})}})}` is formatted to:
```

syntax/format.go
orlade-anz: Another option would be to make the indent an input so the user can change it, kinda like [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify).

However I wouldn't recommend it at this point. Let's see how this feels to use first.

syntax/std_fmt_test.go
ericzhang6222: The behavior of `pretty` is printing the content. If it returns/evals to a value, it will duplicate the content which looks like:
```
(
        c: 3,
        a: 1,
        b: 2
)
(a: 1, b: 2, c: 3)
```
which is not good. So it returns a `""` which can avoid this issue.

syntax/format.go
ericzhang6222: Changed tab to two spaces, and current result looks like:
```
@> //fmt.pretty({'a':1, 'b':2, 'c':(d:11, e:22, f:{111, 222})})
{
  b: 2,
  a: 1,
  c: (
    d: 11,
    f: {
      111,
      222
    },
    e: 22
  )
}
```

syntax/std_fmt_test.go
orlade-anz: Mm, okay, but I don't think it's very helpful to test that a void function doesn't return anything.

Can you capture the string as it's output and assert that it matches the pretty format?

syntax/format_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFmtPrettyForDict(t *testing.T) { //nolint:dupl
+	t.Parallel()
+	simpleDict, err := EvaluateExpr(".", `{'a':1, 'b':2, 'c':3}`)
+	assert.Nil(t, err)
+	str := FormatString(simpleDict, 0)
+	assert.Equal(t, true, contains(str, "\n%sa: 1", 1))
```
orlade-anz, line 16: Nit: replace all `assert.Equal(t, true` with `assert.True(t`

orlade-anz: Actually, the `contains` checks don't seem like a good fit for what you're trying to test. I'd expect to see something of the form:

```
str := FormatString(simpleDict, 0)
assert.Equal(t, `{
  a: 1,
  b: 2,
  c: 3
}`, str)
```
Does that not work for some reason?

syntax/format.go
```diff
@@ -0,0 +1,117 @@
...
+	} else if index == 0 && length > 1 {
+		format = format[1:]
+	} else if index == length-1 && length > 1 {
+		format = format + "\n"
+	}
+
+	return format
+}
+
+func getIdents(identsNum int) string {
```
marcelocantos, line 108: indents, not idents.

marcelocantos:
(In a separate PR) relations should be formatted as tables, e.g.: for a "simple" table: `{|a, b| (1, 2), (12, 13)}`:

```arrai
{|a , b |
 (1 , 2 ),
 (13, 14),
}
```

For a "complex" table, revert to set-of-tuples representation:

```arrai
{
  (
    a: {1, 2},
    b: {3, 4},
  ),
  (
    a: {20, 30},
    b: {40, 50},
  ),
}
```

syntax/format_test.go
ericzhang6222: This can't work as the evaluation result is not order guaranteed. `EvaluateExpr(".", '(a:1, b:2, c:3)')` can't output 
```
{
  a: 1, 
  b: 2, 
  c: 3
}
```
always.

syntax/std_fmt_test.go
ericzhang6222: Yes, the formatted string has been tested in format_test.go and this test case test 3 data types, `dict`, `tuple` and `set` with calling of `//fmt.pretty`.

syntax/format_test.go
orlade-anz: That seems like a bug... Marcelo's guidance has generally been avoid nondeterminism where possible, since it makes everything more complicated.

In this case, it would be quite desirable for a pretty printer to produce deterministic output, especially deterministically sorted output. The items in tuples, dicts and sets should all be ordered in the pretty output, a) so we can test more easily, and b) because it's easier for a user to read.

Is that possible?

ericzhang6222:
> Simple structures should be inlined. E.g., `[1, 2, 3]` should format as is.
> 
> There's no hard and fast rule for what constitutes "simple". As a bare minimum, inline any array or set (up to some threshold length) with only trivial elements (numbers, strings, byte arrays, `()`, `{}` and `true`).
> 
> The threshold could be fixed, or maybe a function of the current indent level.

Will do it in a separated PR. As @nofun97 want a MVP to help debugging ASAP, so I didn't do in this PR.

syntax/format_test.go
```diff
@@ -1,77 +1,67 @@
...
-	simpleDict, err := EvaluateExpr(".", `{'a':1, 'b':2, 'c':3}`)
+	simpleDict, err := EvaluateExpr(".", `{'c':3, 'a':1, 'b':2}`)
 	assert.Nil(t, err)
-	str := FormatString(simpleDict, 0)
-	assert.True(t, contains(str, "\n%sa: 1", 1))
-	assert.True(t, contains(str, "\n%sb: 2", 1))
-	assert.True(t, contains(str, "\n%sc: 3", 1))
+	str, err := FormatString(simpleDict, 0)
+	assert.Nil(t, err)
+	assert.Equal(t, "{\n  a: 1,\n  b: 2,\n  c: 3\n}", str)
```
orlade-anz, line 15: Much nicer :). Nit: Does it still work if you format the expected string like this (much easier to read):
```suggestion
	assert.Equal(t, `
{
  a: 1,
  b: 2,
  c: 3
}`, str)
```

PR #525, opened 2020-07-28
Fix precedence of safe-accessor tail

### LINKED ISSUE #552: Fix precedence of safe-accessor tail

### Description

See #525 for details.

### Steps to Reproduce
1. 
2. 
3. 

### Your Environment
```sh
$ arrai info

```

marcelocantos:
Fixes #552 

Change safe-accessor tail to expect an expr instead of the next rule on the precedence stack.

The expression `a.b?:c.d` used to parse as `(a.b?:c).d`. This is confusing because it means that `.d`, which looks like it belongs to `c`, actually applies to `a.b` if `a` has a `b` attribute. This PR changes it to parse as `a.b?:(c.d)`.

This is actually a bit of a hack. My preference would be to loop back to the current rule. However, wbnf doesn't have a way to refer to the current rule, only the next rule (`@`) or back to the top (`expr`). It might be better to add this capability to wbnf. Comments welcome.

Checklist:
- [x] Added related tests


orlade-anz:
Can you add a description of what this is actually doing for posterity?

orlade-anz (changes requested):
The description is good but hard to grok; can you also add a test case that would have failed before but should pass now and vice versa?

marcelocantos:
We really should fix the stack operator in wbnf so that this rule can be self-referential. https://github.com/arr-ai/wbnf/issues/85

PR #535, opened 2020-08-04
Fix issues with pretty printing

### LINKED ISSUE #516: //fmt.pretty doesn't output valid arr.ai syntax

### Description

Where possible, `//fmt.pretty` should output valid syntax. When pretty-printing dicts with string keys,
the function outputs bare strings, which means the output can't be copy-pasted back into arr.ai. It
can also lead to some weird behaviour (albeit under some pathological edge cases).

### Steps to Reproduce

```arrai
@> //fmt.pretty({'a: 1, b':1})
{
  a: 1, b: 1
}
{'a: 1, b': 1}
@> //fmt.pretty({'42': 1, 42:2})
{
  42: 2,
  42: 1
}
{42: 2, '42': 1}
```

### Expected behavior

Output values as valid arr.ai syntax.

### Actual behavior

Bare strings as keys.

### Your Environment
```sh
$ arrai info
Version    : 65402d624e76a7af10a9bf7e4eb0e8e627526d01 (from v0.96.0)
Git commit : 65402d624e76a7af10a9bf7e4eb0e8e627526d01
Date       : 2020-07-21T10:10:57Z
OS/arch    : linux/amd64
Go version : go1.14.3 linux/amd64
```

orlade-anz:
Fixes #516, part of https://github.com/arr-ai/arrai/issues/171.

- Wrap dict keys in quotes
- Wrap and indent arrays
- Return pretty string instead of logging

Checklist:
- [x] Added related tests


syntax/std_fmt.go
```diff
@@ -15,8 +13,7 @@ func stdFmt() rel.Attr {
 				return nil, err
 			}
 
-			fmt.Println(prettifiedString)
-			return value, nil
+			return rel.NewString([]rune(prettifiedString)), nil
```
ericzhang6222, line 16: Now its output is:
```
@> //fmt.pretty([1,2,3])
'[\n  1,\n  2,\n  3\n]'
@>  
```
It is not prettified.

orlade-anz:
> ```
> @> //fmt.pretty([1,2,3])
> '[\n  1,\n  2,\n  3\n]'
> @>
> ```
> 
> Current output is not right.

It's not perfect, but I'd argue it's more correct that it was. Simple lists should be flat, but in practice, most lists are not simple. Expanding simple lists is a better interim than flattening non-simple lists as we do today.

ericzhang6222:
> > ```
> > @> //fmt.pretty([1,2,3])
> > '[\n  1,\n  2,\n  3\n]'
> > @>
> > ```
> > 
> > 
> > Current output is not right.
> 
> It's not perfect, but I'd argue it's more correct that it was. Simple lists should be flat, but in practice, most lists are not simple. Expanding simple lists is a better interim than flattening non-simple lists as we do today.

I think the issue is code `return rel.NewString([]rune(prettifiedString)), nil` in std_fmt.go.
All `\n` is transformed to char `\n` and will not start a new line really.

marcelocantos:
This is correct behaviour. The purpose of `//fmt.pretty` is to return a string with the correctly formatted content, not to print that string. The intended behaviour will manifest with:

```bash
arrai eval '//fmt.pretty([1,2,3])'
```

orlade-anz:
> Some test cases are failed.

Fixed I believe.

syntax/std_fmt.go
orlade-anz: That's expected per Marcelo's comment.

PR #557, opened 2020-08-09
fix #550, add more concise tuple projection syntax

### LINKED ISSUE #550: More concise tuple projection syntax.

We could do with some concise syntax for simple projection over tuples. Currently, `(:t.x, :t.y)` or `t -> (:.x, :.y)` does the job, but this is a little noisy and there is no way to indicate exclusion.

`t.|x, y|` to project over `x` and `y` (equivalent to `t -> (:.x, :.y)`) and `t.~|x, y|` to project over all attributes except `x` and `y` (no equivalent currently). Both cases fail if the tuple is missing any of the listed attributes.

A equivalent for relations isn't required, since it can be composed as `r => .|x, y|` and `r => .~|x, y|`.

For consistency, this should also be allowed inline in tuple syntax: `(:t.~|x,y|, z: 42)`. A key difference from the otherwise equivalent `t.~|x,y| +> (z: 42)` is that the inline form would fail if `t.~|x,y|` has a `z` attribute.

#### Notes

1. The following change avoids the risk of failing on duplicate attributes: `(t.~|x,y,z|, z: 42)`. However, it will in turn fail if `t` doesn't have a `z` attribute. To perform a conditional override of `z` if present, simply use the `+>` operator, which is explicitly designed for this.
1. Some thought should be give to offering a similar syntax for dicts. Since accessing a single attribute is done via `d('x')`, this should be extended. A fairly obvious extension seems apropos: `d(|'x'|)`.

nofun97:
Fixes #550.

1. Allows `t.|a, b, c|` which takes attributes of `a, b, c`
2. Allows `t.~|a, b, c|` which takes all the attributes of `t` except
`a, b, c`
3. Changes `Names.String` to use `OrderedNames` for deterministic output

Checklist:
- [x] Added related tests

### DIFF
--- a/syntax/expr_tuple_project_test.go
+++ b/syntax/expr_tuple_project_test.go
@@ -1,0 +1,27 @@
+package syntax
+
+import "testing"
+
+func TestTupleProjectExpr(t *testing.T) {
+	t.Parallel()
+
+	AssertCodesEvalToSameValue(t, `(a: 1)            `, `(a: 1, b: 2, c: 3).|a|       `)
+	AssertCodesEvalToSameValue(t, `(a: 1, b: 2)      `, `(a: 1, b: 2, c: 3).|a, b|    `)
+	AssertCodesEvalToSameValue(t, `(a: 1, b: 2, c: 3)`, `(a: 1, b: 2, c: 3).|a, b, c| `)
+	AssertCodesEvalToSameValue(t, `(b: 2, c: 3)      `, `(a: 1, b: 2, c: 3).~|a|      `)
+	AssertCodesEvalToSameValue(t, `(c: 3)            `, `(a: 1, b: 2, c: 3).~|a, b|   `)
+	AssertCodesEvalToSameValue(t, `()                `, `(a: 1, b: 2, c: 3).~|a, b, c|`)
+}
+
+func TestTupleProjectExprError(t *testing.T) {
+	t.Parallel()
+
+	AssertCodeErrors(t, `lhs does not evaluate to tuple: {1: 1}`, `{1: 1}.|a|                      `)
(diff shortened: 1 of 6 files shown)


syntax/expr_tuple_project_test.go
```diff
@@ -0,0 +1,27 @@
...
+	t.Parallel()
+
+	AssertCodeErrors(t, `lhs does not evaluate to tuple: {1: 1}`, `{1: 1}.|a|                      `)
+	AssertCodeErrors(t, `lhs does not evaluate to tuple: {1: 1}`, `{1: 1}.~|a|                     `)
+	AssertCodeErrors(t, `names are not subset of lhs: |a, b, c|`, `(a: 1, b: 2, c: 3).|d|          `)
+	AssertCodeErrors(t, `names are not subset of lhs: |a, b, c|`, `(a: 1, b: 2, c: 3).|a, b, d|    `)
+	AssertCodeErrors(t, `names are not subset of lhs: |a, b, c|`, `(a: 1, b: 2, c: 3).|a, b, c, d| `)
+	AssertCodeErrors(t, `names are not subset of lhs: |a, b, c|`, `(a: 1, b: 2, c: 3).~|d|         `)
+	AssertCodeErrors(t, `names are not subset of lhs: |a, b, c|`, `(a: 1, b: 2, c: 3).~|a, b, d|   `)
+	AssertCodeErrors(t, `names are not subset of lhs: |a, b, c|`, `(a: 1, b: 2, c: 3).~|a, b, c, d|`)
```
marcelocantos, line 26: Support and test for `(a: 1, b:2) -> .|a|` and `(a: 1, b:2) -> .~|a|`.

PR #559, opened 2020-08-09
Optimize nest for single attribute

### LINKED ISSUE #452: Optimize nest for single-attribute behaviour.
The plan was originally for `X nest a` to be a shortcut for `X nest |a|a`, but a better alternative is for the short form `X nest a` to store a set of matching `a` elements rather than a sub-relation with just an `a` attribute:

Normal form:
```arrai
@> {|x,y| (1, 2), (1, 3), (2, 4)} nest |y|y
{(x: 1, y: {(y: 2), (y: 3)}), (x: 2, y: {(y: 4)})}
```

Short form:
```
@> {|x,y| (1, 2), (1, 3), (2, 4)} nest |y|y
{(x: 1, y: {2, 3}), (x: 2, y: {4})}
```

Concerns:

1. How to offer an `unnest` counterpart? There's no obvious way to distinguish syntactically.
    - Perhaps the only option is to have a different operator pair (`group`/`ungroup`?) for the single-attribute case.

nofun97:
Change proposed in this pull request:
- Introduce optimized single attribute nest `x nest y`

Checklist:
- [x] Added related tests

### DIFF
--- a/syntax/expr_single_nest_test.go
+++ b/syntax/expr_single_nest_test.go
@@ -1,0 +1,24 @@
+package syntax
+
+import "testing"
+
+func TestSingleAttrNest(t *testing.T) {
+	t.Parallel()
+
+	AssertCodesEvalToSameValue(t,
+		`{(x: 1, y: {2, 3}), (x: 2, y: {4})}`,
+		`{|x,y| (1, 2), (1, 3), (2, 4)} nest y`,
+	)
+
+	AssertCodesEvalToSameValue(t,
+		`{(x: 1, y: {2, 3, 4})}`,
+		`{|x,y| (1, 2), (1, 3), (1, 4)} nest y`,
+	)
+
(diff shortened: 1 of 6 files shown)


syntax/expr_single_nest_test.go
```diff
@@ -0,0 +1,24 @@
...
+		`{|x,y| (1, 2), (1, 3), (1, 4)} nest y`,
+	)
+
+	AssertCodesEvalToSameValue(t,
+		`{(x: 1, y: {2}), (x: 2, y: {3}), (x: 3, y: {4})}`,
+		`{|x,y| (1, 2), (2, 3), (3, 4)} nest y`,
+	)
+
+	AssertCodesEvalToSameValue(t, `{}`, `{} nest y`)
+}
```
marcelocantos, line 24: Some negative tests would be good.

PR #568, opened 2020-08-11
Add simple unit testing framework for arr.ai

orlade-anz:
Arr.ai is currently largely without a unit testing story. Regardless of the sophistication of the tooling, it's important that devs are at least able to encode expectations of their logic in a standard format. This PR proposes a most basic standard for implementing unit tests, upon which many enhancements are anticipated.

The essence is similar to `go test ./...`: running `arrai test` will walk the current directory tree and collect files named `*_test.arrai`. Each of these files is evaluated (working from its own directory) to produce an arr.ai value, as all arr.ai files do. The data structure produced will be a tree with some number of leaf nodes; all leaf nodes are expected to be `true` (not just truthy); if any leaves are not `true`, the path to that leaf will be reported as a failure.

With this pattern, a "traditional" style of test spec can be written as follows:

```arrai
let (:foo, ...) = //{./foo};

let testFoo =
    foo(2) = 4;

let testFooBar =
    foo('bar') = 'barbar';

(
    :testFoo,
    :testFooBar,
)
```

However it is also possible to creates many different structures of test, or for example import another set of tests and transform them in some way to produce a new set of tests.


ericzhang6222 (approved):
LGTM

PR #585, opened 2020-08-19
fixes extra whitespace after computed empty string

nofun97:
initially the expression
```
$`
    abc
    ${''}
    def
`
```

would output
```
    abc

    def
```

this commit suppresses any lines which only contains computed empty string
```
    abc
    def
```
Checklist:
- [x] Added related tests

### DIFF
--- a/syntax/compile_xstr.go
+++ b/syntax/compile_xstr.go
@@ -180,9 +180,12 @@ func cleanEmptyVal(values rel.Value) []rel.Value {
 		// cleans bare string after the empty computed string
 		cleanRE(firstIndentRE, i+1, func(match, toReplace string) string {
 			if match != "" {
-				// clean bare string before the empty computed string
-				// only cleans this if i+1 will be changed, this is to retain
-				// any whitespaces in the bare string of i-1. Meant to handle
+				// cleans bare string before the empty computed string
+				//
+				// only does this if i+1 will be changed, this is meant to retain
+				// any whitespaces in the bare string of arr[i-1].
+				//
+				// Meant to handle
 				// $`
 				//   abc
 				//   ${''}def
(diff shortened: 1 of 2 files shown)


syntax/compile_xstr.go
```diff
@@ -156,3 +158,53 @@ func xstrConcat(seq rel.Value) (rel.Value, error) {
 	}
 	return rel.NewString([]rune(sb.String())), nil
 }
+
+// this function cleans whitespaces of bare strings before and after a computed emptyt string
```
orlade-anz, line 162: Nit
```suggestion
// cleanEmptyVal cleans whitespaces of bare strings before and after a computed empty string.
```

syntax/compile_xstr.go
```diff
@@ -156,3 +158,53 @@ func xstrConcat(seq rel.Value) (rel.Value, error) {
...
+func cleanEmptyVal(values rel.Value) []rel.Value {
```
orlade-anz, line 163: Nit: If the input must be a rel.Array, I'd be inclined to make the parameter a rel.Array. Otherwise you're just hiding information for no benefit.

syntax/compile_xstr.go
```diff
@@ -156,3 +158,53 @@ func xstrConcat(seq rel.Value) (rel.Value, error) {
...
+	arr := values.(rel.Array).Values()
+	length := len(arr)
+	cleanRE := func(re *regexp.Regexp, index int, cleaner func(string, string) string) {
+		if index >= 0 && index < length {
```
orlade-anz, line 167: Nit: return early if this is _not_ the case; save the reader an indent.

orlade-anz (changes requested):

anzdaddy (approved):

Issue #584, opened 2020-08-18
`arrai compile` to build self-contained executable arr.ai binaries


orlade-anz:
Please **do not post any internal, closed source snippets** on this public issue tracker!

### Purpose

> Please describe the _end goal_ you are trying to achieve that has led you to request this feature.

Currently running arr.ai scripts requires clearing several hurdles: having the right version of `arrai` installed, all source scripts available locally, running form the appropriate working directory, etc. This is a barrier to adoption.

In an ideal world, arr.ai scripts could be compiled to binaries that could be invoked from any context on any platform.

### Suggested approaches
> What have you tried, and how might this problem be solved?

There are multiple possible approaches of increasing sophistication, value and effort. We should probably do each in turn.

1. [P0] Resolve imports and copy all necessary sources into a ZIP file. Allow `arrai run` to handle these bundles by unzipping them to a virtual file system.
2. [P0] Generate an executable Go program containing the bytes of the ZIP file that when executed performs a virtual `arrai run` as above.
3 [P1] Parse and compile the target arr.ai source into an Expr tree. Serialize the Expr tree into Go code that reconstructs the tree. Generate an executable Go program that deserializes and evals that tree.

There are further possible refinements of course, but they would be relatively insignificant.

PR #590, opened 2020-08-20
add context to all eval functions

nofun97:
Part of #584 

- Adds context to all eval functions

The context var will be used to contain filesystem which can be used to compile arrai programs

**note to reviewer**: I put a few `TODO` and `FIXME`, please read those before approving


nofun97:
the function `WrapContext` is changed to `WrapContextErr` to avoid confusion

marcelocantos (approved):

PR #600, opened 2020-08-26
Add interpretation of generic structs to NewValue

orlade-anz:


orlade-anz:
Now supported via `rel.NewValue`.

PR #616, opened 2020-08-31
Add contrib/json.arrai for working with JSON data

orlade-anz:
The standard, canonical JSON structure is a pain to work with. In most cases the decoded data need not be reversible, and follows a schema that makes a tuple more appropriate than a map. This PR adds functions to apply these simplifications.


contrib/json.arrai
```diff
@@ -0,0 +1,38 @@
...
+# Returns a "simplified" structure, replacing decoded tuples with their internal values.
+# Intended for cases in which the types of null values is not interesting.
+let rec simplify = \j
+    cond j {
+        [...]: j >> simplify(.),
+        (:a): simplify(a),
+        (:s): simplify(s),
+        (): {},
+        _: cond {
+            j < 99999999999999999999: j,
```
nofun97, line 12: is this meant to check if `j` is a number?

contrib/json.arrai
```diff
@@ -0,0 +1,38 @@
+# Helper functions for working with JSON data.
+
+# Returns a "simplified" structure, replacing decoded tuples with their internal values.
+# Intended for cases in which the types of null values is not interesting.
+let rec simplify = \j
```
nofun97, line 5: also handle `(:b)` for boolean values.

nofun97 (changes requested):

contrib/json.arrai
orlade-anz: Yeah. I forget now what the problem was, but I couldn't figure out how else to do it.

orlade-anz: Added a comment as well.

PR #631, opened 2020-09-05
Print all failed test in a test suite which is an arr.ai tuple.

ericzhang6222:
Fixes # .

- Print all failed test cases in an arr.ai test file
```
$ ./arrai test
Tests:
contrib/json_test.arrai
contrib/util_test.arrai
contrib/xml_test.arrai
examples/test/multiple_cases_test.arrai
examples/test/single_case_test.arrai

failed test:
not equal
expected: Hello, arrai!
actual:   Hello22, arrai!

examples/test/multiple_cases_test.arrai:4:18:
    # Added expression let to clarify the actual result can be a complex expression
    testCase1: //test.assert.equal("Hello, arrai!", let words = ["Hello22", "arrai!"];//seq.join(", ", words)),

unexpected size
expected: 2
actual:   {222, Hello, arrai!}

examples/test/multiple_cases_test.arrai:6:18:
    testCase2: //test.assert.unequal("Hello, arrai!", //seq.join(", ", ["Nihao", "arrai!"])),
    testCase3: //test.assert.size(2, {"Hello", "arrai!", "222"}),
examples/test/multiple_cases_test.arrai failed
4/5 tests passed
```

Checklist:
- [x] Added related tests


rel/expr.go
```diff
@@ -86,3 +87,24 @@ func EvalExpr(ctx context.Context, expr Expr, local Scope) (_ Value, err error)
...
+// Errors is sequence of error
+type Errors struct {
+	errors []error
+}
+
+// Error returns string represents Errors
+func (errs Errors) Error() string {
+	var result strings.Builder
+	len := len(errs.errors)
+	if len > 0 {
```
marcelocantos, line 100: Test is redundant. The range loop will short-circuit if `errs.errors` is empty.

rel/expr.go
```diff
@@ -86,3 +87,24 @@ func EvalExpr(ctx context.Context, expr Expr, local Scope) (_ Value, err error)
...
+		for i, err := range errs.errors {
+			result.WriteString(err.Error())
+			if i < len-1 {
+				result.WriteString("\n\n")
+			}
```
marcelocantos, line 105:
```suggestion
			if i > 0 {
				result.WriteString("\n\n")
			}
			result.WriteString(err.Error())
```

rel/expr_tuple.go
```diff
@@ -142,12 +142,18 @@ func (e *TupleExpr) String() string { //nolint:dupl
 func (e *TupleExpr) Eval(ctx context.Context, local Scope) (Value, error) {
 	tuple := EmptyTuple
 	var err error
+	errs := Errors{}
```
marcelocantos, line 145: `var err` (line above) should be declared where it's used, not at the top of the function.

marcelocantos, line 145: Don't modify the core behaviour of `ExprTuple.Eval` for the sake of the testing infrastructure. Reporting multiple errors will break normal error reporting. Implement `EvalTests` instead.

marcelocantos (changes requested):

rel/expr_tuple.go
ericzhang6222: The following loop will use it, so declared here.

ericzhang6222: It is used in following loop with code:
```
	for _, attr := range e.attrs {
		tuple, err = attr.Apply(ctx, local, tuple)
```
I think current code is better, as `tuple` has to be declared in the top of func. Or it has to add code to transfer  looks like:
```
	for _, attr := range e.attrs {
		interTuple, err := attr.Apply(ctx, local, tuple)
                tuple = interTuple
```

rel/expr_tuple.go
```diff
@@ -140,6 +140,10 @@ func (e *TupleExpr) String() string { //nolint:dupl
 
 // Eval returns the subject
 func (e *TupleExpr) Eval(ctx context.Context, local Scope) (Value, error) {
+	if cmd, isStr := ctx.Value(CmdIdentifier{Name: "test"}).(string); isStr && cmd == "test" {
+		return e.evalTests(ctx, local)
+	}
+
```
marcelocantos, line 146: This shouldn't be required. `test.Test` should explicitly unpack the TupleExpr and evaluate each component.

marcelocantos (changes requested):

rel/expr.go
```diff
@@ -86,3 +87,29 @@ func EvalExpr(ctx context.Context, expr Expr, local Scope) (_ Value, err error)
...
+		}
+		result.WriteString(err.Error())
+	}
+	return result.String()
+}
+
+// CmdIdentifier identifies an arrai command like test, shell and eval etc.
+type CmdIdentifier struct {
+	Name string
+}
```
orlade-anz, line 115: This should live in `internal/test/`, since it should only be used in the case of testing.

In general, test-related code should not be mixed in with implementation code.

rel/expr_tuple.go
```diff
@@ -151,3 +155,22 @@ func (e *TupleExpr) Eval(ctx context.Context, local Scope) (Value, error) {
 	// TODO: Construct new tuple directly
 	return tuple.(*GenericTuple).Canonical(), nil
 }
+
+func (e *TupleExpr) evalTests(ctx context.Context, local Scope) (Value, error) {
```
orlade-anz, line 159: As above, this should be in `internal/test`. It could probably be an extension of the `isRecursivelyTrue` function, rather than a separate thing (i.e. in a test, we would want to report all of the errors at all levels of a data structure, not just for tuples and not just one level deep).

orlade-anz (changes requested):

rel/expr_tuple.go
ericzhang6222:
```
type TupleExpr struct {
	ExprScanner
	attrs   []AttrExpr
	attrMap map[string]Expr
}
```
This is current definition of `TupleExpr`, so it has to change `attrs` to `Attrs` to unpack in `test.Test`. Is it acceptable? Or it can add new functions in `expr_tuple.go` or package `rel` for `test`.

rel/expr_tuple.go
ericzhang6222: Current code can't support this approach. The code `value, err := expr.Eval(ctx, scope)` calculates the value, as its implementation when one error happens in evaluation, it will stop executing and return a `nil` as `value`. So it can't calculate the value with some code like `isRecursivelyTrue`.
And as current `expr` definition, it can't do the similar thing like `isRecursivelyTrue` too, refers https://github.com/arr-ai/arrai/pull/631#discussion_r487479789.
I think it has to do a leverage, open access to attributes in `expr` or add some functions in package `rel` which can be used by `test`.

orlade-anz: So my thinking is:

1. The `_test.arrai` file should at least compile. If it doesn't you shouldn't expect good error reporting.
2. If the test compiles, then its output will be a data structure: a set, array, dict or tuple (assume for now it's not just a single bool, though that should be allowed).
3. Consider the data structure as a tree: it has some branches (internal tree-shaped nodes) and some leaves (primitives, probably booleans). For each leaf that is not `true`, we want to report a failure and include the _path_ through the branches to that leaf, the comparison that failed, and what the expected and actual values were.

To do this, we first evaluate the file and get the data structure and walk through all of the nodes. For each leaf node, add an `Entry` to a list with the path we took to get there. If the leaf is not `true`, then we do some more work to complete the report.

I'm going to make a key assumption here because it simplifies things a lot and will be largely true in practice. Assume that the data structure containing the test output is statically defined (i.e. a literal, not generated by some other code). Each branch is either a set, array, dict or tuple, and each leaf is some other kind of expression.

So to get the failed comparison, parse the test file and walk the AST in search of the lowest tuple or sequence attribute found in the failed path. If there are multiple (e.g. `empty` which could be used in different cases), work your way back up through the parent nodes to find the right one.

Once you have the leaf node (unevaluated at this point), search inside that for a comparison operator (something that returns true: `=`, `>`, `<:`, `(<=)`, etc.). When you find one, replace it with some analogue to `//test.assert*` which, instead of returning `true` or `false`, evaluates to a tuple with the data you want to report (expected source and value, actual source and value).

Evaluate that modified AST and walk it, compiling the report from the leaf tuples. Fin.

I expect that I'm glossing over the complexity of "find the comparison", but it's conceptually possible.

You might argue "wow that's hard; this PR is an improvement, shouldn't we just merge it and do the more complete solution as a follow-up?" Maybe. I'm wary because a) it's mixing test code with implementation which is a code smell for both test and implementation code, and b) it's not in the same direction as the solution described above; we would ultimately plan to delete this test code, so why commit it in the first place and add an extra complexity and maintenance burden?

marcelocantos:
Closing as abandoned / superseded.

The useful product goal (report every failing case in a multi-case `*_test.arrai`, especially when cases error rather than returning false) is captured in https://github.com/arr-ai/arrai/issues/725.

This PR’s approach (special-casing `TupleExpr.Eval` via test context) should not be revived; any implementation should live in `pkg/test` only. Boolean multi-case reporting already works there.

PR #635, opened 2020-09-06
move handleEvalOut to its own package

nofun97:
This function is moved into its own package so arr.ai binaries can import this function

### DIFF
--- a/pkg/output/out.go
+++ b/pkg/output/out.go
@@ -1,0 +1,129 @@
+package output
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/arr-ai/arrai/pkg/ctxfs"
+	"github.com/arr-ai/arrai/rel"
+	"github.com/spf13/afero"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/output/out.go
```diff
@@ -0,0 +1,129 @@
+package output
```
marcelocantos, line 1: `pkg/arrai` would be better.

pkg/output/out.go
```diff
@@ -0,0 +1,129 @@
...
+	"path"
+	"strings"
+
+	"github.com/arr-ai/arrai/pkg/ctxfs"
+	"github.com/arr-ai/arrai/rel"
+	"github.com/spf13/afero"
+)
+
+// HandleEvalOut handles output writing for evaluated values.
+func HandleEvalOut(ctx context.Context, value rel.Value, w io.Writer, out string) error {
```
marcelocantos, line 17: This seems to be just outputting, not "handling" output. Perhaps call it `OutputValue`.

marcelocantos (changes requested):

PR #636, opened 2020-09-06
first cut on arrai binary

nofun97:
Part of #584 

> [P0] Generate an executable Go program containing the bytes of the ZIP file that when executed performs a virtual arrai run as above.

Checklist:
- [x] Added related tests


cmd/arrai/build.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/arr-ai/arrai/syntax"
+	"github.com/arr-ai/arrai/tools"
+	"github.com/arr-ai/wbnf/parser"
+	"github.com/spf13/afero"
+	"github.com/urfave/cli/v2"
+)
+
+var compileCommand = &cli.Command{
+	Name:    "compile",
+	Aliases: []string{"compile"},
```
orlade-anz, line 24: Nit: useless?

cmd/arrai/build.go
```diff
@@ -0,0 +1,151 @@
...
+
+func compile(c *cli.Context) error {
+	tools.SetArgs(c)
+	file := c.Args().Get(0)
+
+	return compileFile(arraictx.InitRunCtx(context.Background()), file, c.Value("out").(string))
+}
+
+func compileFile(ctx context.Context, path, out string) error {
+	if err := runFileExists(ctx, path); err != nil {
```
orlade-anz, line 40: Nit: update name?

cmd/arrai/build.go
```diff
@@ -0,0 +1,151 @@
...
+		return err
+	}
+
+	bundledScripts := bytes.Buffer{}
+	if err := bundleFiles(ctx, path, &bundledScripts, ""); err != nil {
```
orlade-anz, line 45: Nit: passing an empty string as an arg is a bit of a smell. Maybe refactor `bundleFiles` so that `bundleFiles` handles this simple case, and there's some other `bundleFilesTo` that takes a non-empty `out` param? Something like that.

internal/build/main.arrai
```diff
@@ -0,0 +1,26 @@
+\bytes
```
orlade-anz, line 1: Nit: It would be nice to have a comment in this file to explain what it's for. It's pretty trippy.

cmd/arrai/build.go
```diff
@@ -0,0 +1,151 @@
...
+	defer f.Close()
+
+	if err := buildBinary(ctx, bundledScripts.Bytes(), f); err != nil {
+		return err
+	}
+	return nil
+}
+
+func createGoFile(ctx context.Context, bundledScripts []byte) ([]byte, error) {
+	template, err := Asset("internal/build/main.arrai")
```
orlade-anz, line 66: Nit: Should this really be using arr.ai? It looks like a very simple Go template would do the job with much less fuss?

orlade-anz (changes requested):

cmd/arrai/build.go
nofun97: I just thought of using arrai and I didn't really know how to use go template 😅

orlade-anz: Give it a shot I reckon; it would save us from all of the go-bindata cruft, and probably even be less source code as well. Not sure how to encode the bytes, but possible something like `NewNumber(b).Repr()` is an option.

cmd/arrai/compile_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+
+	"github.com/arr-ai/arrai/pkg/arraictx"
+	"github.com/arr-ai/arrai/pkg/ctxfs"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCompileFile(t *testing.T) {
```
orlade-anz, line 18: Can you add a test with an import? Local and/or remote? Perhaps some of the same things that Chloe's import tests use (e.g. https://github.com/ChloePlanet/arrai-examples)

cmd/arrai/compile_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func testExec(t *testing.T, path, expected string, args ...string) {
+	path, err := filepath.Abs(path)
+	assert.NoError(t, err)
+	c := exec.Command(path, args...)
+	actual := bytes.Buffer{}
+	c.Stdout = &actual
+
+	assert.NoError(t, c.Run())
```
orlade-anz, line 64: Nit: I think you can just use `c.Output()` for a nicer API.

orlade-anz (changes requested):

cmd/arrai/compile_test.go
nofun97: don't think it's necessary, since the imports are handled with the `bundle` functions and those functions are already tested.

cmd/arrai/run_test.go
```diff
@@ -86,7 +86,7 @@ func TestRunBundle(t *testing.T) {
 			ctx := ctxfs.SourceFsOnto(context.Background(), getImportFs(t))
 			ctx = ctxrootcache.WithRootCache(ctx)
 			zipped := &bytes.Buffer{}
-			require.NoError(t, bundleFiles(ctx, syntax.MustAbs(t, c.filePath), zipped, ""))
+			require.NoError(t, bundleFilesTo(ctx, syntax.MustAbs(t, c.filePath), zipped, ""))
```
orlade-anz, line 89: nit: `bundleFiles`, remove the `""`

orlade-anz (changes requested):

cmd/arrai/compile.go
```diff
@@ -0,0 +1,173 @@
...
+
+    if err := arrai.OutputValue(ctx, value, os.Stdout, ""); err != nil {
+        panic(err)
+    }
+}
+`
+
+type bin struct {
+	Data string
+}
```
orlade-anz, line 52: Nit: I think this can be inlined if it's only used once, like 
```
struct{ Data string }{Data: strings.Join(data, ",")}
```

cmd/arrai/compile_test.go
orlade-anz: Fair enough.

orlade-anz (approved):

PR #731, opened 2026-08-28
Fix -cpuprofile/-memprofile defer scope

marcelocantos:
## Summary
- `prepareProfilers()` registered cleanup defers inside itself, so `-cpuprofile`/`-memprofile` stopped immediately after setup and captured almost nothing.
- Return a cleanup closure that `main` defers (and runs before `os.Exit` paths, including `--help-agent`).

## Test plan
- [x] `arrai -cpuprofile=/tmp/c.prof eval 1+1` then `go tool pprof -top /tmp/c.prof` shows real samples
- [ ] Spot-check `-memprofile` writes a heap profile at exit

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/arrai/main.go
+++ b/cmd/arrai/main.go
@@ -37,7 +37,8 @@ var cmds = []*cli.Command{
 }
 
 func main() {
-	prepareProfilers()
+	stopProfilers := prepareProfilers()
+	defer stopProfilers()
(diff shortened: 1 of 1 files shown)


marcelocantos:
Unblocks profiling for #733.

PR #732, opened 2026-08-28
Honor go.mod-pinned versions (and replace) for module imports

marcelocantos:
## Summary
- Restores go.mod version pinning for `//{github.com/...}` module imports, which regressed when `anz-bank/pkg/mod` was replaced with always-`@latest` download during the frozen generics migration.
- Prefer the effective `Dir` from `go list -m -json all` so **`replace` directives** are honoured (matches old pkg/mod behaviour more closely than re-downloading `path@version`).
- Resolve the graph relative to the **importing script's module root**, not process cwd.
- Tests cover pinned version, `replace => ./local`, module-root≠cwd, longest-prefix match, fallback to `@latest`, and invalid paths.

## Test plan
- [x] `go test ./syntax/ -run 'RetrieveModule|RequiredModule'`
- [ ] Manual: project with `replace github.com/foo => ./local` importing `//{github.com/foo/...}` resolves to local
- [ ] Manual: `arrai run` from outside the module still sees that module's pins

Made with [Cursor](https://cursor.com)

### DIFF
--- a/syntax/compile.go
+++ b/syntax/compile.go
@@ -1335,7 +1335,7 @@ func (pc ParseContext) compilePackage(ctx context.Context, b ast.Branch, c ast.C
 			}
 			return NewImportExpr(scanner, expr, importPath), nil
 		}
-		expr, err := importExternalContent(ctx, pkg.Scanner(), decoderTuple, name)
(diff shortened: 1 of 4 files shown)


marcelocantos:
Correctness fix called out from #733 (repro had to vendor because imports ignored go.mod pins).

PR #736, opened 2026-08-30
Harden go.mod pin resolution for module imports

AriehSchneier:
## Summary
- `retrieveModule`'s go.mod-pin discovery (`go list -m -json all`) fails outright when `go.sum` lacks an entry for a required module, silently falling back to `@latest` — pass `-mod=mod` so it self-heals `go.sum` the same way `go mod download` does.
- When a pin *is* found but downloading that exact version fails, stop silently degrading to `@latest`; return an error instead, since a pin should mean "use this version or fail."
- Documents the resulting versioning behavior in `docs/docs/lang/import.md`, filling in a long-standing "TODO: versioning" placeholder under Remote Imports.

Follow-up to #732, which restored go.mod-pin resolution for module imports after it regressed during the `anz-bank/pkg` removal.

## Test plan
- [x] `go test ./syntax/...`
- [x] `golangci-lint run ./syntax/...`
- [x] New tests reproduce each bug when its fix is reverted (verified manually), and pass with the fix applied

### DIFF
--- a/docs/docs/lang/import.md
+++ b/docs/docs/lang/import.md
@@ -89,7 +89,25 @@ repository. This can be done by the following:
 //{github.com/username/repo_name/path/to/script}
 ```
 
-TODO: versioning
+#### Versioning
+
+By default, `arrai` fetches the latest available version of the repository.
+
+If the importing script's project has a `go.mod` (found the same way as for
+Package Imports above) that already `require`s the module being imported, that
+pinned version is used instead — including following a `replace` directive
+that points it at a fork or a local filepath. This lets you pin or override
+remote imports the same way you would a regular Go dependency:
+
+```go
+require github.com/username/repo_name v1.2.3
+
+replace github.com/username/repo_name => ../local/fork
+```
+
+It is not possible to pin a version inline in the import path itself (e.g.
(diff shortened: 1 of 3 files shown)


AriehSchneier:
One behavioral difference from pre-v0.321 (anz-bank/pkg/mod) that this PR deliberately doesn't address, flagging for visibility:

Old code mutated the consuming project's `go.mod`/`go.sum` on first use of an *undeclared* remote import (via `go get`), which permanently pinned it after that first resolution — later invocations would find it now declared and reuse that exact version.

Current code (`downloadModule`, using `go mod download` rather than `go get`) never writes to `go.mod`. An undeclared remote import is re-resolved to whatever's `@latest` on *every* invocation, with no persistent record — so two runs on different days could silently resolve an undeclared import to different commits, with nothing in the repo showing that happened.

This is arguably a reasonable trade-off (no surprise side-effects on a user's `go.mod` from what looks like a read-only `arrai` invocation), but it's a real reproducibility regression from pre-v0.321 for anyone relying on undeclared/implicit remote imports, so it should be a conscious decision rather than an implicit side-effect of this refactor. Happy to open a follow-up issue if that's preferred over discussing here.

marcelocantos (approved):

PR #737, opened 2026-08-30
Fix profiler cleanup on interrupt and Fatal error paths

AriehSchneier:
## Summary

Follow-up to #731, which fixed `prepareProfilers()`'s cleanup defers firing immediately at setup instead of at program exit. That fix covered the normal `app.Run()` error-return path and the `-help-agent` early exit, but two more exit paths still bypassed it entirely — `os.Exit` (which `Fatal` calls trigger) skips every deferred function in the process, not just the current stack frame's:

1. **No signal handling.** `serve`/`sync` are long-running commands normally stopped with Ctrl-C. With no `SIGINT`/`SIGTERM` handler, that hits the default disposition and kills the process without running any defers — so `-cpuprofile`/`-memprofile` still captured nothing for what's likely the most common real use case (profiling a running server). Verified before/after: sending `SIGINT` to a profiled `serve` process on `master` left a 0-byte, unparseable pprof file (and needed `SIGKILL` to actually terminate); it now exits promptly with a valid profile.
2. **Existing `Fatal` calls in command actions.** `serve`, `serve_grpc`, and `sync` called `log.Fatal`/`logrus.Fatal` directly on error paths reachable during normal execution (bind failures, bad TLS config, watch failures). These are all inside functions with an already-established path back to `app.Run()`'s error return, so converting them to returned errors costs nothing and lets `main()`'s existing (correct) cleanup path handle it.

`stopProfilers()` is now idempotent (atomic-guarded) since the signal handler can race with `main()`'s own deferred/explicit calls to it.

## Test plan
- [x] `go build ./...`, `go test ./...`, `golangci-lint run ./...`
- [x] Manually verified: `SIGINT` during a profiled `serve` run now flushes a valid CPU profile (previously 0 bytes, process didn't even terminate on `SIGINT`)
- [x] Manually verified: a profiled `serve` bind-conflict error and a profiled `json` parse error both now return normally and flush a valid profile

### DIFF
--- a/cmd/arrai/json.go
+++ b/cmd/arrai/json.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 
-	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 
 	"github.com/arr-ai/arrai/pkg/fu"
@@ -32,7 +31,7 @@ func fromJSON(cli *cli.Context) error {
 	}
 	val, err := translate.StrictTranslator().ToArrai(data)
 	if err != nil {
-		logrus.Fatal(err)
+		return err
 	}
 	fmt.Println(fu.Repr(val))
 	return nil
--- a/cmd/arrai/main.go
+++ b/cmd/arrai/main.go
@@ -5,10 +5,13 @@ import (
 	"fmt"
 	"log"
 	"os"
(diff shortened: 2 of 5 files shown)


marcelocantos (approved):

PR #738, opened 2026-08-31
Fix Relation/Dict Hash128 disagreeing with their own Equal

AriehSchneier:
## Summary

Three confirmed hash/equals contract violations introduced by #734's single-pass hashing rewrite. All three break the fundamental invariant that Equal values must hash the same, so a hash-based Set/Map can silently fail to find or dedupe an existing entry depending on internal representation, construction order, or the specific values involved:

- **`Relation.Hash128`** hashed rows in raw positional order, but `EqualRelation` canonicalizes rows into sorted-attribute-name order before comparing. Two `Equal` relations with different internal attribute ordering (e.g. produced by different join/projection code paths) could hash differently.
- **`Dict.Hash128`** mixed in a `dictSalt` that `GenericSet`/`EmptySet` don't have, even though `Dict.Equal` deliberately treats a `Dict` as equal to a generic `Set` of the same `(@, @value)` entries (and an empty `Dict` as equal to `EmptySet{}`). Equal `Dict`/`Set` values could hash differently.
- **`Array.Hash128`** combined per-index item hashes with `xor`, but each item hash already `xor`s together an index-hash and a value-hash via `hashTuple2`. When the same value occurs at two different indices (e.g. `["x", "x"]`), the value-dependent part of those two item hashes is identical and cancels out under `xor`, making the array's hash completely independent of the repeated value — `["x", "x"]` and `["y", "y"]` hashed identically despite being unequal. Fixed by combining item hashes with `Mix` instead, which isn't self-cancelling like `xor`.

Each is pinned down with a minimal, isolated unit test.

## Context

Found while investigating a real-world regression: a downstream project's script that dedups statement calls and relocates their annotations started non-deterministically dropping annotations after upgrading past `f897d42`. Bisected by individually reverting each of that commit's five named sub-changes (indexed `where`, `Reduce`→`MapBuilder`, `let rec`, direct calls, `GenericTuple` hash caching) against the real failing workload — none alone explained it — which pointed at the remaining "single-pass hashing" surface. Reading each type's `Hash128()` against its own `Equal()` turned up the `Relation` and `Dict` issues; tracing intermediate values through the actual reproduction with `//log.print` showed the annotation relation was already incomplete immediately after `sysl.newNormalize(sysl.load(proto))`, which led to `Array.Hash128` and the exact `["AppA", "AppA"]`/`["AppB", "AppB"]` collision.

All three together fully resolve the original regression — verified against the real production workload (previously non-deterministic annotation loss, now clean across repeated runs).

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite passes)
- [x] `golangci-lint run ./...` (0 issues)
- [x] New unit tests fail on `f897d42`/pre-fix code, pass after

### DIFF
--- a/rel/hash128.go
+++ b/rel/hash128.go
@@ -28,7 +28,6 @@ var (
 	valuesSalt   = hash128.String("rel.Values")
 	emptySetSalt = hash128.String("rel.EmptySet")
 	arraySalt    = hash128.String("rel.Array")
-	dictSalt     = hash128.String("rel.Dict")
 	unionSalt    = hash128.String("rel.UnionSet")
 	relationSalt = hash128.String("rel.Relation")
 
--- a/rel/value_set_array.go
+++ b/rel/value_set_array.go
@@ -120,12 +120,19 @@ func (a Array) Hash(seed uintptr) uintptr {
 	return a.Hash128().Seeded(seed)
 }
 
-// Hash128 computes the 128-bit hash of an Array: the xor of its item tuples.
+// Hash128 computes the 128-bit hash of an Array by mixing its item tuples in
+// index order. Each item's hash already binds its index (via hashTuple2), so
+// this doesn't need to be order-independent like a set/dict combination —
+// and using Mix here, rather than Xor, matters: xor-ing per-element hashes
+// together is vulnerable to cancellation when the same value occurs at two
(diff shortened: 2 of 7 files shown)


marcelocantos (approved):

PR #739, opened 2026-08-31
Evaluator performance overhaul: 42x end-to-end, 106x on eval

marcelocantos:
The 2026-08 evaluator performance overhaul. Arieh's sysl reconstruct workload (in-tree as `perf/reconstruct`, output byte-identical to v0.321.0 throughout) goes from 58.8–67.8s as reported on v0.338.0 to **1.51s** — **42× end-to-end, 106× on the evaluation phase alone** (0.93s of the remainder is compiling the program and sysl library, untouched by this PR). A handcrafted C++ implementation of the same scenario (`perf/reconstruct/native`, byte-identical, CI-checked) puts the native floor at 6.3ms. Full narrative with per-commit measurements: `docs/perf-ledger-2026-08.html`.

Highlights, in landing order:

- Scopes as lexical frames with cached identifier addresses; single-binding frames stored inline
- Tuples as shape-backed structs; relation rows shared with tuples zero-copy
- Relations in a flat arena with selection-vector views and a lazy hash index
- Parallel evaluation of `where`, `=>`, `>>` (arrays) and `>>`/`>>>` (dicts) closure bodies — one knob shared with frozen (`FROZEN_CONCURRENCY` / `frozen.SetMinParallelChunk`), deterministic first-element errors
- Strings: linear concatenation with ownership-tracked append buffers, ASCII byte backing (4× smaller payloads), interned literals, `$"..."` compiled to a dedicated expression rendering into a byte builder
- Bounded memos: `//grammar.parse` results, compiled grammars, regexps
- Reflection-free sorts (`slices.SortFunc` sweep); `rel.All` (`iter.Seq`) with a measured adoption doctrine
- The Hash128/Equal contract reimplemented for the arena representation on top of #738, with its regression tests ported and passing
- `ARRAI_TIMING=1` reports compile and eval phases separately
- Testing: a `slowpath` build tag forces every fast path to its fallback across the whole corpus (differential oracle); `perf/reconstruct` pins output byte-for-byte against a v0.321.0 golden with a two-way allocation ratchet (44M → 37.5M over the PR); the native C++ implementation is byte-checked in CI
- Updates frozen to v1.14.0

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01C5kAkbw85pHGg9S5DUEYGA


AriehSchneier:
Found a correctness regression while bumping [anz-bank/sysl](https://github.com/anz-bank/sysl)'s arrai dependency across this branch.

**Symptom:** sysl's OpenAPI importer (`.arrai` scripts) silently drops path parameters. A relation built via `reduce`+tuple-merge (`acc +> (path |: {paramObject})`) and then mapped with `=>` into a name-keyed dict comes out empty, so any lookup against it fails with `NoReturnError`.

**Bisected to:** 32d51ea "Implement S1–S5 of evaluation-as-query-execution (🎯T24)" — first bad commit; its parent (b3d67e3) is good.

**Confirmed not a downstream (sysl) code issue:** the exact same, unmodified `.arrai` source produces correct output under `v0.321.0` and incorrect output under this branch's tip (78cf1ab, which includes the compiled-plan nil-scanner fix from earlier in this PR). Not a single-element-relation edge case — reproduces with both 1 and 2 path params.

**Minimal repro** (run directly against source, no bundling needed):

`repro.yaml`:
```yaml
openapi: "3.0.2"
info:
  title: Repro API
  version: "1.0"
paths:
  "/persons/{uuid}":
    post:
      operationId: "Enrol_image"
      parameters:
        - in: path
          name: uuid
          schema:
            type: string
          required: true
      responses:
        "201":
          description: OK
```

From an anz-bank/sysl checkout:
```
$ arrai run pkg/importer/openapi/import_openapi_cli.arrai -- --app-name Test --input repro.yaml
```

- `v0.321.0`: correctly renders `/persons/{uuid <: string}:`
- current branch tip: `Call: no return values for input 'uuid' from set {}`

Bisection command used (from an arrai checkout, run against the sysl script above):
```
git bisect start <bad> v0.321.0
git bisect run <script that greps run output for "uuid <: string">
```

Happy to share the full bisect script/logs if useful. Given the size of 32d51ea, didn't attempt to pin the exact line — flagging here rather than guessing further.

AriehSchneier:
Design note while fixing the `Array.Equal`/`seqPipeline` asymmetry a couple commits up: `GenericSet.Equal` and `Relation.Equal` have the same narrow single-type-switch pattern that caused that bug (only matching their own concrete type, rather than a generic `case Set:` fallback like `Dict.Equal` already uses).

I tried several concrete triggers to see if it's reachable (empty relation vs `EmptySet`, a mixed-shape set filtered down to one homogeneous shape, generic-set vs relation literals) and all came back symmetric — arr.ai auto-promotes same-shaped tuple sets to their canonical type at construction, and there's no lazy "pipeline" wrapper type for `GenericSet`/`Relation` the way `seqPipeline`/`dictPipeline` exist for `Array`/`Dict`, so the asymmetry has no way to surface today.

Not changing anything since I can't demonstrate it's currently exploitable, but flagging in case a future lazy wrapper type gets added for either — `GenericSet.Equal`/`Relation.Equal` (and `String.Equal`/`Bytes.Equal`, same pattern) could adopt `Dict.Equal`'s more robust style pre-emptively.

AriehSchneier:
While extending the arrai-native plan-roundtrip test corpus (`pkg/test/testdata/planroundtrip/`), came across what looks like dead code: `rel.NewWildcardExpr` (`rel/expr_tuple.go`) has zero call sites anywhere in the codebase, including `syntax/compile.go`.

The related mechanism it seems to exist to support — an `AttrExpr` with `name == "*"`, auto-detected in `rel.NewAttrExpr` when the attr's value is a `*DotExpr` whose `Attr() == "*"` (i.e. surface form `expr.*`) — is also unreachable from the current grammar. The `get` production (`syntax/arrai.wbnf`) only admits `"." ("&"? IDENT | STR | "~"? names)` after a dot; there's no alternative for a literal `*`. Confirmed empirically: `t.*` fails to parse with `unconsumed input` right after `t`.

Not filing this as a bug since there's no working syntax that reaches it to demonstrate broken behavior — flagging in case it's meant to back a feature that never got wired up to the grammar (or is safe to remove).

AriehSchneier:
Pushed 18 commits adding an arrai-native plan-roundtrip test corpus (`pkg/test/testdata/planroundtrip/*_test.arrai`), plus a Go harness (`pkg/test/plan_roundtrip_test.go`) that runs every case both directly and through `LowerPlan`/`EncodePlan`/`DecodePlan`. Goal: let coverage for the compiled-plan feature grow as plain arr.ai boolean assertions instead of a hand-maintained Go table.

That corpus immediately found and fixed 6 more bugs (all with regression tests, stash-verified to fail without the fix):

- Plan decode (`rel.compareCtor`) only recognized 8 of 20 comparison operators — the whole subset/superset family (`(<)`, `(>)`, `(<=)`, `(>=)`, `(<>)`, `(<>=)` and their `!`-negated forms) errored "unknown compare op" inside a compiled bundle. Unified into one `rel.CompareOps` table shared by compile-time and plan-decode.
- `median` silently ran as `min` when plan-decoded — `NewMedianExpr`'s `ReduceExpr` format string said `"%s min ???"` (copy-paste typo), and plan lowering identifies which reduce op it is by substring-matching that string.
- The `unnest` keyword has never worked in any version — `compileArrow`'s `"unnest"` case was a bare `panic("unfinished")`, even though `rel.NewUnnestExpr`/`Unnest` and plan lower/lift support were fully implemented and just never wired up. Two-line fix.
- A constant-folded all-numeric bytes literal (e.g. `<<97, 98>>`) lost its source scanner (`NewBytesExpr` returned a bare `NewBytes(...)` instead of `NewLiteralExpr(scanner, ...)` like every other literal fold), breaking comparisons against equivalent string-form bytes literals.
- `(...t, b: 2)` / `{...d, 'b': 2}` crashed the compiler — the tuple/dict grammar's `pairs` rule shares its `extra` (`...`) alternative with the pattern grammar, but the expression-side compilers never checked for it. No merge-via-spread semantic exists to implement instead (see the `NewWildcardExpr` dead-code comment above), so this is now a clean compile error rather than a panic.
- `->*` ("touch") syntax also panicked (`panic("unfinished")`) with no implementation anywhere to wire up (unlike `unnest`) — same treatment, now a clean "not implemented" error.

All confirmed pre-existing since v0.321.0 except the first two (compare-op table gap, median), which are specific to this branch's plan.bin feature. Full test suite green throughout.

AriehSchneier:
Since the last update: pushed 36 more commits continuing to grow the plan-roundtrip corpus (`pkg/test/testdata/planroundtrip/`), now covering coverage-guided sweeps of the stdlib packages (`//encoding.json`/`.yaml`/`.csv`, `//archive`, `//re`, `//test`, `//bits`, `//rel`, `//fmt`, `//os`, `//seq`, `//eval`) and much deeper exercise of this branch's positional-relation internals (parallel `where`/`>>`, `projectDots`'s injective-share-store vs dedup paths, and the group-by cache that reuses a parent view's index for a derived child view). Corpus statement coverage of `rel`+`syntax` combined: 43.5%→50.4%.

Two more instances of the same scanner-loss bug class flagged before (a bare `Value` returned as `Expr` without `NewLiteralExpr`, so it inherits whatever generic/default `Source()` its own type falls back to instead of its real file position):

- `compileBytes`'s empty-literal (`<<>>`) branch had the exact same bug as `NewBytesExpr`'s fold path (fixed earlier) — a second call site that skipped the wrap.
- `compileMacro`'s result — a `{: macro : content :}` embed's value, evaluated eagerly at parse time via `SetCall` against the macro's `@transform` — was returned directly as the `Expr`, even though a real scanner (`MacroValue.Scanner()`) was sitting right there unused.

And a real, independent bug found via the same coverage sweep — **`Concatenate`'s generic fallback silently corrupts data** for anything other than two contiguous zero-offset `String`/`Array` pairs (i.e. `Bytes` concatenation, which has no fast path at all, or an array with holes/a non-zero offset, which fails the fast path's shape check). It used `a.Count()` as the position to start placing `b`'s elements, which is only correct when `a`'s `"@"` indices are gapless and start at 0:

```
[1, , 3] ++ [4, 5]      => [1, , 4, 5]           # the 3 silently overwritten/dropped
5\[1, 2, 3] ++ [4, 5]   => 3\[4, 5, 1, 2, 3]      # fully scrambled
```

Fixed by tracking one past `a`'s highest actual `"@"` index while enumerating it (already being done in the same loop), instead of assuming count == next index.

All three fixes have stash-verified regression tests (confirmed to fail without the fix, pass with it). Full test suite green throughout (`go test ./... -count=1`).

AriehSchneier:
Since the last update: pushed 10 more commits continuing the plan-roundtrip corpus (`pkg/test/testdata/planroundtrip/`). Corpus statement coverage of `rel`+`syntax` combined: 50.4%→53.7%.

One real bug found and fixed, with a stash-verified regression test:

- **`GenericSet.ArrayEnumerator` panicked on any element that wasn't an `"@"`-indexed tuple.** Its ordering comparator did an unchecked `a.(Tuple).MustGet("@")`, assuming every element was array-item-shaped — an assumption that can never actually hold for a genuine `GenericSet`, since a homogeneous set of `"@"`-indexed tuples canonicalises to a specialised type (`Array`/`Dict`/`String`/`Bytes`) instead of staying a `GenericSet`. Any real `GenericSet` therefore always holds a mix of shapes, and enumerating it in array order (e.g. via `//log.printf`, one of the few real callers of `ArrayEnumerator`) crashed with `interface conversion: ... is not rel.Tuple: missing method Enumerator`:
  ```
  {[1, 2, 3] | {(x: 1, y: 2)}, 99}   # a UnionSet + a Number both land in the
                                      # generic bucket, forming a genuine
                                      # GenericSet - enumerating it in array
                                      # order used to panic
  ```
  Fixed by ordering via the general `Value` total order instead (delegating to the already-correct `OrderedValues`), matching every other "no natural position" fallback in this codebase.

Also fixed a test-quality issue in the existing corpus: `seqPipelineParallel` compared `(//seq.repeat(600, [0]) >> . + 1)` against the *literal same expression* on both sides, which is tautological — a systematic bug in `force`'s parallel range-splitting (e.g. positions scrambled across chunk boundaries) could corrupt both sides identically and still compare equal. Rebuilt both sides from genuinely different computations that must still agree.

Coverage highlights from this round: `UnionSet` (formed by `|`-ing sets that land in different "buckets", e.g. an `Array` with a differently-shaped `Relation`) had **zero coverage anywhere in the corpus** despite being a real, easily-reachable code path — now at 75-100% across `Has`/`With`/`Without`/`Where`/`CallAll`/`Kind`/`Less`/`Negate`/enumeration. Also rounded out `pushWhereThroughProject` (the `rel => (x: .x) where .x = 5` → `(rel where .x = 5) => (x: .x)` compile-time rewrite) from 26.3%→94.7% — its one remaining branch is provably dead: an explicit `\pattern body` lambda's body always recursively swallows a trailing `where` (confirmed via the grammar), so `where` can only ever bind outside a `=>` via the implicit-dot shorthand, which is always an ident pattern.

Full test suite green throughout (`go test ./... -count=1`).

PR #740, opened 2026-08-31
Broaden Hash128/Equal regression coverage for #738's fixes

AriehSchneier:
## Summary

Follow-up to #738 (merged). Three additions closing test-coverage gaps the original fix didn't cover, no production code changes:

- **`Array.Hash128`'s cancellation test only covered a two-element array with an adjacent repeat.** Extended to non-adjacent repeats, an all-equal array (with an *even* element count — an odd count doesn't fully cancel under the old `xor` formula, since xor-ing the same term three times leaves one copy uncancelled, so a 3-element all-equal case would misleadingly pass even against the buggy code), and a repeat nested one level down.
- **Added a test asserting each specialized tuple kind (`StringCharTuple`, `BytesByteTuple`, `ArrayItemTuple`, `DictEntryTuple`) hashes the same as its `GenericTuple` equivalent.** `GenericTuple.Equal` is permissive across kinds (unlike each specialized kind's own, narrower `Equal`), so this is exactly the same class of bug as `Relation`/`Dict.Hash128` from #738, and wasn't covered by any existing test.
- **Added `FuzzArrayHash128RepeatedValue`**, generalizing the repeated-value cancellation case across arbitrary values and repeat counts (2–7) instead of a handful of hand-picked cases. In the process this surfaced a bug in the *test harness itself* (not production code): invalid UTF-8 byte sequences collapse to the same replacement rune through `[]rune` conversion, so two distinct fuzzer-provided byte strings can be genuinely equal once round-tripped through `NewString` — the fuzz target now compares post-round-trip. Seeded with the input that surfaced this as a regression case. 6M+ generated executions found no further issues.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (full suite passes)
- [x] `golangci-lint run ./...` (0 issues)
- [x] Reverting `Array.Hash128`'s fix locally confirms both the table test and the fuzz corpus catch the original bug again
- [x] `go test -fuzz=FuzzArrayHash128RepeatedValue -fuzztime=30s` — 6M+ executions, no failures against the current fix

### DIFF
--- a/rel/testdata/fuzz/FuzzArrayHash128RepeatedValue/cd6bfdb1fe13f4b0
+++ b/rel/testdata/fuzz/FuzzArrayHash128RepeatedValue/cd6bfdb1fe13f4b0
@@ -1,0 +1,4 @@
+go test fuzz v1
+string("\x89")
+string("\xb8")
+byte('R')
--- a/rel/value_set_array_test.go
+++ b/rel/value_set_array_test.go
@@ -20,18 +20,90 @@ import (
 // completely independent of the repeated value -- so, for example,
 // ["AppA", "AppA"] and ["AppB", "AppB"] hashed identically despite being
 // unequal, silently corrupting any hash-based Set/Map containing such
-// arrays as elements or as part of a composite key (this is how it was
-// found: architecture-specs' fully-qualified app name paths often repeat
-// a segment, e.g. `MVISION :: MVISION`).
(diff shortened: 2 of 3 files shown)


marcelocantos (approved):
Reviewed: test-only, cases verified locally (including the fuzz seeds), and the go directive is 1.25 so the parallel range-var subtests are sound. CI is green despite the earlier concern. Merging.

PR #742, opened 2026-08-31
Resolve go.mod pins against the top-level project for nested imports

AriehSchneier:
## Summary

- `importModuleFile` resolved a go.mod pin by walking up from the *importing file's own* directory (`findRootFromModule`). For a top-level script that's the user's project, but for a nested import (a file pulled in as a dependency, itself doing a `//{github.com/...}` import) that directory is inside the read-only Go module cache. `go list -mod=mod -m -json all` there fails outright trying to self-heal `go.sum` (`permission denied`), so the pin lookup silently comes back empty and resolution falls back to slow `@latest` git resolution — every time, for every nested import.
- Real-world impact: `arrai bundle`/`arrai r` on a script with a couple of nested imports went from ~0.7s (pre-pin-resolution `v0.321.0`) to 45-66s.
- Fix: resolve every import in a run against a single "primary" module root, computed once from the top-level project and reused for nested imports regardless of which file actually contains the `import` statement. This matches Go's actual module resolution — one resolved version per module via minimal version selection over the *main* module's go.mod — rather than treating each dependency's own go.mod as authoritative.
- While tracing this, also found that `go list -mod=mod`'s go.sum self-heal (intentional, for a go.sum missing entries not needed to build the main module) was left mutating the caller's real `go.sum`/`go.mod` in place — a real behavior difference from `v0.321.0`, which never touched either file. Fixed by snapshotting both files before resolving and restoring them afterward, so import resolution never leaves a project's checked-in files changed.

## Test plan

- [x] New unit tests for the primary-root memoization contract (`pkg/ctxrootcache/ctxrootcache_test.go`)
- [x] New integration-style test proving the exact regression: a "nested" directory's own (different) go.mod pin no longer wins over the top-level project's pin (`TestPrimaryRootMakesNestedImportsUseTheTopLevelPin`)
- [x] New tests proving `go.mod`/`go.sum` come out byte-identical after resolution, including the self-heal-from-nothing case (`TestLoadRequiredModulesDoesNotLeaveGoSumChanged`, `TestLoadRequiredModulesPreservesExistingGoModAndGoSum`)
- [x] `go build ./...`, `go vet ./...`, `go test ./...`, `golangci-lint run` all clean
- [x] Verified against a real production workload (nested imports through multiple modules): 45-66s → ~0.7-1.4s, byte-identical output to both the pre-regression baseline and `v0.321.0`, and zero `go.sum`/`go.mod` diff left behind (previously 65 lines of unrelated churn)

### DIFF
--- a/pkg/arraictx/ctx.go
+++ b/pkg/arraictx/ctx.go
@@ -32,6 +32,7 @@ func InitRunCtx(ctx context.Context) context.Context {
 	ctx = ctxfs.SourceFsOnto(ctx, afero.NewOsFs())
 	ctx = ctxfs.RuntimeFsOnto(ctx, afero.NewOsFs())
 	ctx = ctxrootcache.WithRootCache(ctx)
+	ctx = ctxrootcache.WithPrimaryRootCache(ctx)
 	ctx = buildinfo.WithPackageBuildData(ctx)
 	return ctx
 }
--- a/pkg/ctxrootcache/ctxrootcache.go
+++ b/pkg/ctxrootcache/ctxrootcache.go
@@ -48,3 +48,46 @@ func rootCacheFrom(ctx context.Context) *sync.Map {
 	}
 	return m.(*sync.Map)
 }
+
+type primaryRootKeyType int
+
+const primaryRootKey primaryRootKeyType = iota
+
+// primaryRoot memoizes the module root used to resolve go.mod-pinned module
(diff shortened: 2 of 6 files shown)


marcelocantos (approved):
Reviewed: primary-root resolution matches Go's MVS model, the snapshot/restore keeps checked-in files untouched, and the regression tests pin both behaviours. Merging.

PR #743, opened 2026-09-01
Document and surface -cpuprofile/-memprofile flags

AriehSchneier:
## Summary
- `-cpuprofile`/`-memprofile` were entirely undiscoverable: `prepareProfilers()` strips them from `os.Args` by hand before `cli.App` is built, so they were never registered as `cli.Flag`s and never showed up in `--help` for any of the three binary personalities (`arrai`, `ai`, `ax`).
- Adds them to `app.Flags` purely for `--help`/usage display — actual parsing is unchanged, still handled by `prepareProfilers()` since profiling has to start before `app.Run()`.
- Adds a Command Line docs page (`docs/docs/cli/profiling.md`) covering usage, the leading-flag/`=`-syntax requirement, and how they interact with long-running commands (see #737).

## Test plan
- [x] `go build ./...`, `go test ./cmd/...`, `golangci-lint run ./cmd/...`
- [x] Manually verified `--help` output lists both flags for `arrai` and `ai`
- [x] Manually verified `-cpuprofile=<path>` still works end-to-end after the flag declarations were added

### DIFF
--- a/cmd/arrai/main.go
+++ b/cmd/arrai/main.go
@@ -23,6 +23,20 @@ import (
 //go:embed agents-guide.md
 var agentsGuide string
 
+// profilerFlags document -cpuprofile/-memprofile for `--help`. They're never
+// actually parsed by cli: prepareProfilers() strips them out of os.Args
+// before the app is built, since profiling must start before app.Run().
+var profilerFlags = []cli.Flag{
+	&cli.StringFlag{
+		Name:  "cpuprofile",
+		Usage: "Write a CPU profile to `FILE` for the duration of the run. Must precede the command name.",
+	},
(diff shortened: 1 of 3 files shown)


marcelocantos (approved):
Looks good. Flags were previously invisible in `--help` because `prepareProfilers()` strips them before `cli.App` is built; declaring them on `app.Flags` for display only is the right split, and the docs page matches the actual `-cpuprofile=`/`-memprofile=` parser.

Nit, not blocking: urfave/cli v2 will still render `--cpuprofile value` (space form, double dash). `prepareProfilers` only consumes leading `-cpuprofile=` / `-memprofile=`. The new docs page is the source of truth.

Issue #12, opened 2020-01-15
Ascii charts??


marcelocantos:
Might be useful to support https://github.com/guptarohit/asciigraph in arr.ai.

PR #744, opened 2026-09-01
Drop the anz-bank/afero fork; fix real afero's MemMapFs quirks

AriehSchneier:
## Summary

- Removes `replace github.com/spf13/afero => github.com/anz-bank/afero v1.2.4`, a fix from 2020 (#12, "temporary fix to allow sysl to use arrai in go") that tagged a personal fork pointing at an unmerged spf13/afero PR ([spf13/afero#226](https://github.com/spf13/afero/pull/226)), fixing `MemMapFs` not resolving Windows volume names. That PR was never merged and the underlying issue ([spf13/afero#225](https://github.com/spf13/afero/issues/225)) is still open upstream — but the fork also silently pinned the entire dependency ~80 releases behind on every platform, not just Windows.
- Bumps to real upstream `spf13/afero v1.15.0` (latest).
- Removing the fork surfaced a second, more general incompatibility: real `afero.MemMapFs` (still true in v1.15.0) treats a bare relative path (`"a.arrai"`) and its root-anchored form (`"/a.arrai"`) as distinct, unrelated entries, unlike a real OS filesystem. arrai's relative-import resolution and `--out` writer both build bare relative paths and expect root-nesting, so this broke widely once the fork was removed. Verified via a minimal, non-mocked repro against real afero (write "file.txt", `afero.Walk` never finds it, `afero.ReadFile` on the walked path panics).
- Adds `ctxfs.NewTestMemMapFs()` — a small wrapper that normalizes every path (root-anchoring bare paths, stripping Windows volume names) before it reaches `MemMapFs` — and switches `CreateTestMemMapFs` and the handful of tests using a raw `afero.NewMemMapFs()` to it.

Checked for documentation referencing the fork or these afero behaviors (CLAUDE.md, README, dev docs, code comments) — nothing existed to update; all explanatory context lives in the new Go doc comments.

## Test plan
- [x] `go build ./...`
- [x] `go test -race -tags timingsensitive ./...`
- [x] `golangci-lint run ./...`
- [x] New unit tests for the path-normalizing wrapper (`pkg/ctxfs/rooted_memmapfs_test.go`)


marcelocantos (approved):
Good cleanup. The 2020 `anz-bank/afero` replace was pinning every platform ~80 releases behind for a Windows MemMapFs volume-name bug that is still open upstream. The in-repo `rootedMemMapFs` shim is the right replacement: it implements the full `afero.Fs` surface, covers the relative-vs-rooted and volume-name cases, and tests lock the contract.

`CreateTestMemMapFs` dropping `filepath.Abs` is correct once the wrapper roots bare paths. The `Compile` `filepath.Rel` change is cosmetic-label only and correctly no longer fails compilation.

Leftover raw `afero.NewMemMapFs()` in `TestRunBundleWithoutRoot` is bundle-path based and CI-green; not a merge blocker.

PR #747, opened 2026-09-02
go.mod pin resolution: preserve mtime, prefer readonly, auto-add unpinned imports

AriehSchneier:
## Summary

Six related fixes/improvements to the go.mod-pin-resolution path (`syntax/gomod.go`, from #732/#736), found while root-causing a real `.arraiz` rebuild-flakiness issue in a consuming project's CI, then digging into the feature's actual history.

**1. Preserve go.mod/go.sum mtime when restoring after self-heal.** #742's `restore()` writes the original content back with `os.WriteFile`, which bumps the mtime to "now" even when the bytes are byte-identical to what was there before — so every resolution touching go.mod aged it forward, cascading into spurious `.arraiz` rebuilds downstream (confirmed via `make --debug=v`: `Prerequisite 'scripts/go.mod' is newer than target '...'`, despite content never changing). Fixed by skipping the rewrite when content already matches, and restoring the original mtime via `os.Chtimes` when a real rewrite is needed.

**2. Try `-mod=readonly` before falling back to the self-heal path.** Traced the actual history: `v0.321.0` (via `anz-bank/pkg/mod`, removed in `deb7e85`) resolved imports using `go mod download` (bare) + `go list -m` (no `all`) — lighter queries that never needed self-healing. `deb7e85` replaced that with a naive `@latest`-only resolver, and #732/#736 later re-added pin-awareness using `go list -m -json all`, which needs the *full* unpruned graph and routinely needs `-mod=mod`'s self-heal to get it, even when go.sum is otherwise complete. Verified against a real project: `-mod=readonly` succeeds outright (zero errors, same ~0.02s) whenever go.sum is already sufficient — the common case. Only retry with `-mod=mod` (still snapshot/restore-protected) when readonly actually fails. This means the self-heal/restore path — and the TOCTOU race a concurrent `arrai bundle` could hit racing another process over go.mod/go.sum — no longer executes at all for the common case.

**3. Add unpinned imports to go.mod instead of just downloading them.** Also restores `v0.321.0` behaviour: `anz-bank/pkg@v0.0.48`'s `Get()` ran `go get <path>` for an import that wasn't already resolvable, persisting a require directive (and go.sum entry) — confirmed by reading its actual source. The `deb7e85` replacement (carried through #732/#736) only ever ran `go mod download`, which populates the module cache but never touches go.mod — so a never-declared import stayed on the slow, unpinned path forever, on every run, with no way to become fast short of a human manually running `go get`. `retrieveModule`'s fallback now calls a new `addModule` (`go get`, run in `moduleRoot`) instead of `downloadModule` when go.mod doesn't already require the import; a successful add evicts the memoized module graph so a later resolution in the same run sees it immediately. `downloadModule` itself is unchanged, still used for the already-pinned-but-not-cached case, which must never touch go.mod.

**4. Stop shortening past a confirmed module boundary.** `addModule`'s fallback loop chops the last path segment and retries on any `go get` failure — but when `go get` reports `module X found (version), but does not contain package Y`, `X` *is* the correct, confirmed module (the failure is that the specific file isn't in the resolved version, e.g. a stale proxy cache). Continuing to shorten past a confirmed boundary risks landing on a different, wrong (too-shallow) module that happens to independently exist — reproduced for real against a multi-module monorepo. Fixed by recognizing this specific error and retrying once with the confirmed path directly, cutting a 5-segment path down to at most 2 attempts (82s → 40s in the reproduction).

**5. Skip pinning with no module context at all.** `moduleRoot` can legitimately be `""` — the shell/REPL has no source file to derive one from, or `findRootFromModule` walks to the filesystem root without finding a `go.mod`. Previously this meant `addModule` always tried `go get` with `cmd.Dir` unset: if the process's cwd happens to be inside some unrelated project, that silently pollutes its go.mod; if cwd isn't inside any module at all, `go get` hard-fails (`'go get' is no longer supported outside a module`). Fixed by checking `go env GOMOD` first and falling back to the plain, unpinned `downloadModule` path (which works fine module-less) when there's no ambient module to pin into.

**6. Stop the shortening loop on transient `go get` errors.** Generalizes fix 4: a network/proxy/checksum failure (`dial tcp`, `no such host`, timeouts, connection refused/reset, checksum mismatch, `SECURITY ERROR`) says nothing about whether the attempted path is a real module, so treating it the same as "no such module" and shortening can mask the real error or land on a different, wrong-but-real shorter module. `addModule` also now snapshots/restores go.mod around the entire attempt (not just the `go get` failure case), so a `go get` that succeeds but whose immediately-following `downloadModule` call fails doesn't leave an unconfirmed `require` behind.

## Test plan

- [x] `TestLoadRequiredModulesPreservesGoSumMtime`, `TestRetrieveModuleFallsBackWithoutPin` (extended), `TestRetrieveModulePersistsAcrossRuns`: each fails against its respective pre-fix code and passes with it (verified via `git stash` for each).
- [x] Fix 2: verified against a real project's `go.mod`: `-mod=readonly` succeeds with zero errors, identical timing to `-mod=mod`; full existing suite (pin resolution, replace-dir, self-heal fallback) unchanged.
- [x] Fix 3: verified end-to-end: resolving a brand new import added correct `go.mod`/`go.sum` entries (matching real `go get` output), and a second run resolved the same import in 0.14s via the fast pinned path with zero further file changes.
- [x] Fix 4: `TestRetrieveModuleStopsAtModuleBoundaryOnMissingPackage` (uses `cloud.google.com/go`/`cloud.google.com/go/pubsub`, a real public multi-module monorepo resolved via the public Go proxy — no private-repo network access needed, so it also runs cleanly in this repo's own public CI).
- [x] Fix 5: `TestRetrieveModuleWithoutAmbientModuleFallsBackToDownload` (chdir into a module-less temp dir, assert success); confirmed via `git stash` to fail with the exact `'go get' is no longer supported outside a module` error pre-fix.
- [x] Fix 6: `TestIsTransientModuleErr` (table-driven classifier test) and `TestRetrieveModuleStopsImmediatelyOnTransientError` (poisons `GOPROXY`, asserts the loop attempts the full path exactly once instead of shortening through every prefix); both confirmed via `git stash` to fail against pre-fix code.
- [x] End-to-end against a real consuming project's data throughout: nested-import resolution via a script needing a private, nested Go dependency (itself requiring `arr-ai/arrai`) still correct and fast (0.68-1.1s, 80-85MB peak RSS) with zero unwanted go.mod/go.sum changes when nothing needs adding.
- [x] `go build ./...`, `go vet ./...`, `go test ./...`, `golangci-lint run` all clean.

Happy to split fix 3 into its own PR if that's easier to review separately — it's a larger behavioural change than # 1/# 2 (go.mod now gets a real, persistent write for a genuinely new dependency, by design) — let me know.

Note: this PR's own `Lint` (`make check-clean`) check is currently failing, but it's pre-existing and unrelated — `master`'s tip is missing a "regenerate embedded stdlib bundles" commit after its last dependency bump, and the same failure reproduces on other unrelated open PRs.

Also includes an unrelated small CI fix: `TestRetrieveModuleStopsImmediatelyOnTransientError` asserted on Unix-specific "connection refused" wording and failed on `windows-latest` (Windows phrases a refused dial as `connectex: ... actively refused it` instead — a test bug, not a product one, since `isTransientModuleErr` already matched via the OS-independent `dial tcp` branch either way). Fixed the assertion, and added `fail-fast: false` to the Test job's OS matrix so one platform failing doesn't cancel the others mid-run.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -54,6 +54,7 @@ jobs:
   test:
     name: Test
     strategy:
+      fail-fast: false
       matrix:
         platform: ${{ vars.BUILD_AND_TEST_OS_MATRIX && fromJSON(vars.BUILD_AND_TEST_OS_MATRIX) || fromJSON('[ "ubuntu-latest", "macOS-latest", "windows-latest" ]') }}
     runs-on: ${{ matrix.platform }}
--- a/syntax/gomod.go
+++ b/syntax/gomod.go
@@ -246,6 +246,12 @@ func retrieveModule(importPath, version, moduleRoot string) (*goModule, error) {
 		}
 	}
 
+	// moduleRoot's go.mod doesn't already require this: add it there with
+	// `go get` so this and every later resolution of it -- in this run, and
+	// any future one -- becomes a fast, pinned lookup instead of hitting
+	// this slow, unpinned path again. (This restores v0.321.0's behaviour,
+	// lost when anz-bank/pkg/mod -- whose Get() did the same `go get` call
+	// -- was replaced by an early, unpinned version of this file.)
 	modPath := importPath
 	var lastErr error
 	for {
@@ -253,8 +259,12 @@ func retrieveModule(importPath, version, moduleRoot string) (*goModule, error) {
 		if len(parts) < 3 {
 			break
 		}
-		m, err := downloadModule(modPath, version)
+		m, err := addModule(modPath, version, moduleRoot)
 		if err == nil {
+			// go.mod just changed; forget any memoized graph for moduleRoot
+			// so a later resolution in this same run sees the addition
+			// instead of re-adding it or missing it.
+			requiredModulesByRoot.Delete(moduleRoot)
 			return m, nil
 		}
 		lastErr = err
@@ -262,15 +272,17 @@ func retrieveModule(importPath, version, moduleRoot string) (*goModule, error) {
 	}
 	if lastErr != nil {
 		if ee, ok := lastErr.(*exec.ExitError); ok {
-			return nil, fmt.Errorf("go mod download %s: %s", importPath, ee.Stderr)
+			return nil, fmt.Errorf("go get %s: %s", importPath, ee.Stderr)
 		}
-		return nil, fmt.Errorf("go mod download %s: %w", importPath, lastErr)
+		return nil, fmt.Errorf("go get %s: %w", importPath, lastErr)
 	}
-	return nil, fmt.Errorf("go mod download %s: module not found", importPath)
+	return nil, fmt.Errorf("go get %s: module not found", importPath)
 }
 
 // downloadModule runs `go mod download -json` for modPath at version (or
 // "latest" if version is empty) and returns the resulting module info.
+// Unlike addModule, this never changes go.mod -- it's only used for a module
+// go.mod already pins but whose content isn't in the local module cache yet.
 func downloadModule(modPath, version string) (*goModule, error) {
 	arg := modPath
 	if version != "" {
@@ -291,4 +303,25 @@ func downloadModule(modPath, version string) (*goModule, error) {
 		return nil, fmt.Errorf("go mod download %s: parsing output: %w", arg, err)
 	}
 	return &goModule{Name: result.Path, Dir: result.Dir}, nil
+}
+
+// addModule runs `go get` for modPath at version (or "latest" if version is
+// empty) in moduleRoot, adding a require directive (and go.sum entry) to its
+// go.mod, then looks up the resulting directory the same way an
+// already-pinned module would be.
+func addModule(modPath, version, moduleRoot string) (*goModule, error) {
+	arg := modPath
+	if version != "" {
+		arg += "@" + version
+	} else {
+		arg += "@latest"
+	}
+	cmd := exec.Command("go", "get", arg) //nolint:gosec
+	if moduleRoot != "" {
+		cmd.Dir = moduleRoot
+	}
+	if _, err := cmd.Output(); err != nil {
+		return nil, err
+	}
+	return downloadModule(modPath, version)
 }
--- a/syntax/gomod.go
+++ b/syntax/gomod.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 )
 
 // goModule holds the result of module resolution (from go list or go mod download).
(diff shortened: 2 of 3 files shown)


AriehSchneier:
Added a 4th fix while investigating a follow-up question about auto-add's behavior against real multi-module dependencies (like architecture-specs' \`acceleration-tools\` monorepo).

**4. Stop shortening the module path once \`go get\` confirms a real boundary.** \`retrieveModule\`'s fallback loop treats every \`go get\` failure the same way -- chop the last segment, retry. But \`go get\`'s error \`module X found (version), but does not contain package Y\` means X *is* the real, confirmed module boundary; the failure is just that the specific file isn't in the resolved version (typically a stale proxy cache serving an older pseudo-version). Continuing to shorten past a confirmed boundary can overshoot into a different, too-shallow module that happens to independently exist -- reproduced for real: resolving a nested import under \`acceleration-tools/infra-diagrams/builder\` intermittently landed on the repo-root \`acceleration-tools\` module instead, causing a hard failure two steps later when the file wasn't there either. Fixed by recognizing this specific error and retrying once with the confirmed path directly. Cuts up to 5 sequential \`go get\` attempts down to at most 2, and eliminates the wrong-boundary failure mode.

\`TestRetrieveModuleStopsAtModuleBoundaryOnMissingPackage\` demonstrates the efficiency win directly (82s -> 40s for the real dependency), though the specific "wrong module" failure mode is tied to proxy cache staleness and isn't reliably reproducible in a fast deterministic test.

marcelocantos (approved):
Reviewed all three fixes plus the workflow change — approving and merging.

**On the red Lint check: it's pre-existing master breakage, not this PR.** Chased it down, since it's the same phenomenon your chat message describes:

- The `.arraiz` bundles embed `module/github.com/arr-ai/arrai/go.mod` verbatim.
- The committed bundles are **stale**: their embedded go.mod still says `grpc v1.79.3` / `protobuf v1.36.10`, while master's go.mod says `v1.82.1` / `v1.36.11` (dependabot #722 bumped go.mod without regenerating the bundles).
- So the bundles differ from their sources *whenever `make` actually decides to rebuild them* — which is exactly the mtime dependency you identified. Checked out cold (CI) or after anything touches go.mod, make rebuilds, the new bundle embeds the current go.mod, and `check-clean` fails.

Verified: `make check-clean` on **plain master** passes with stale mtimes but fails identically the moment you `touch go.mod`; and a docs-only npm dependabot PR fails the same step, which rules out any Go change as the cause. Fix is a regeneration commit on master (`make generate`), and worth considering whether Go dependency bumps should trigger it automatically — every future dependabot Go bump re-breaks this.

Locally on this PR head: `golangci-lint` 0 issues, `go test ./syntax/...` green.

**On the fixes themselves** — 1 and 2 are unambiguous wins, and the `fail-fast: false` matrix change is welcome (we lost a macOS result to a Windows failure recently). The `transientModuleErr` classification in fix 3 is a nice catch: shortening the path on a network blip could silently pin a wrong-but-real monorepo root, which would be a miserable bug to diagnose.

**One durability caveat on fix 3, for the record** (not a blocker — the failure mode is graceful degradation to today's behaviour): a `require` added for an `.arrai`-only dependency is removed by the next `go mod tidy`, since no Go file imports it and there's no `tools.go` trick available when there's no importable Go package. Verified empirically. So the pin sticks until someone tidies, then the slow path returns silently. Worth thinking about whether the durable answer is an arrai-owned sidecar lock file, or an explicit `arrai mod tidy`-style subcommand that scans `.arrai` sources (direct + transitive) and writes the pins deliberately — which would also let the runtime go strictly read-only and just point at the command when it hits an unpinned import.

